PR #6, opened 2019-01-16
`Response.successful` method

### LINKED ISSUE #5: `JSONDecodeError` are not helpful when HTTP status was not successful
Instead, Python Client should raise the HTTP error as an exception.

E.g. if a 404 happens under the hood, don't show `JSONDecodeError`, show the 404

pcattori:
Fixes #5 

Monkey-patched method that
1. raises *useful* exceptions when HTTP errors occur
2. returns the response (for method chaining)

This PR also includes a unit test to ensure the client does raise HTTP errors when they are encountered.

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -70,6 +70,8 @@ class Client:
         :type method: str
         :param endpoint: API endpoint to call (relative to the Base API path for this client).
         :type endpoint: str
+        :return: HTTP response
+        :rtype: :class:`requests.Response`
         """
         url = "/".join([self.origin, self.base_path, endpoint])
         response = requests.request(method, url, auth=self.auth, **kwargs)
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -83,25 +83,41 @@ class Client:
 
         return response
 
-    def get(self, endpoint, **kwargs):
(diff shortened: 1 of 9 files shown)


tamr_unify_client/client.py
```diff
@@ -81,25 +83,41 @@ def request(self, method, endpoint, **kwargs):
 
         return response
 
-    def get(self, endpoint, **kwargs):
-        """Calls :func:`~tamr_unify_client.Client.request` with the ``"GET"`` method.
+    def request_json(self, method, endpoint, **kwargs):
```
nbateshaus: An alternative (and, maybe, cleaner?) approach would be to keep `get()`, but override the `.json()` method of the returned object to call `self.raise_for_status()` before delegating to the default `.json()`.

nbateshaus: It's too bad that `raise_for_status()` doesn't return `self`, so you can't do `client.get(...).raise_for_status().json()`, which would be ideal. BUT, we can inject a method that does `raise_for_status()` then returns `self`. I think that's even better than magical mystery `get()` - add a `successful()` method that does `raise_for_status` and returns `self`, then change call sites to `client.get(...).successful().json()`

pcattori:
This branch was orphaned when we made the repo public on Github. Moved to #7

PR #7, opened 2019-01-16
`Response.successful`

### LINKED ISSUE #5: `JSONDecodeError` are not helpful when HTTP status was not successful

pcattori:
Fixes #5 

...monky-patched `requests.Response` method to:
1. raise HTTP errors as exceptions
2. return the response (for method chaining)

This PR also adds unit tests to ensure the client correctly surfaces HTTP errors as exceptions.

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -70,6 +70,8 @@ class Client:
         :type method: str
         :param endpoint: API endpoint to call (relative to the Base API path for this client).
         :type endpoint: str
+        :return: HTTP response
+        :rtype: :class:`requests.Response`
         """
         url = "/".join([self.origin, self.base_path, endpoint])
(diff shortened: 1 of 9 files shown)


tamr_unify_client/client.py
```diff
@@ -3,6 +3,17 @@
 from tamr_unify_client.models.dataset.collection import DatasetCollection
 from tamr_unify_client.models.project.collection import ProjectCollection
 
+# monkey-patch Response.successful
+from requests import Response
+
+
+def successful(self):
```
nbateshaus: Docstring: "Returns `self` if successful, raises HTTPError otherwise."

PR #32, opened 2019-02-05
Docs for offline installs

pcattori:
via `pip download` and `pip install --no-index--find-links`


pcattori:
@mrazo-tamr made changes. can you take another look?

PR #34, opened 2019-02-13
Add lookup-by-externalID to Dataset/Project. Also add dataset /status

Andy-L:


tamr_unify_client/models/base_collection.py
```diff
@@ -69,3 +69,33 @@ def stream(self, resource_class):
 
     def __iter__(self):
         return self.stream()
+
+    @abstractmethod
+    def by_external_id(self, resource_class, external_id):
+        """Retrieve an item in this collection by relative ID.
```
pcattori: `s/relative/external`

tamr_unify_client/models/base_collection.py
```diff
@@ -69,3 +69,33 @@ def stream(self, resource_class):
...
+
+        Subclasses should override this method and pass in the specific
+        ``resource_class`` for that collection.
+
+        :param resource_class: Resource class corresponding to items in this collection.
+        :type resource_class: Python class
+        :param external_id: The external ID.
+        :type external_id: str
+        :returns: The specified item, if found.
+        :rtype: ``resource_class``
```
pcattori: should include a `:raises KeyError: If <condition when this exception is thrown>`

tamr_unify_client/models/base_collection.py
```diff
@@ -69,3 +69,33 @@ def stream(self, resource_class):
...
+        resources = self.client.get(self.api_path, params=params).successful().json()
+        items = [
+            resource_class.from_json(self.client, resource_json)
+            for resource_json in resources
+        ]
+
+        if len(items) == 0:
+            raise KeyError(f'No item found with external ID "{external_id}"')
+        elif len(items) > 1:
+            raise ValueError(
```
pcattori: My gut instinct is that this should also be a `KeyError`. Usually `ValueError` is for telling the user that the params they passed in were not of the right type / format, whereas `KeyError` indicates that a lookup attempt failed.

Do you think `ValueError` is better than `KeyError` here?

tamr_unify_client/models/dataset/resource.py
```diff
@@ -59,3 +65,14 @@ def records(self):
         with self.client.get(self.api_path + "/records", stream=True) as response:
             for line in response.iter_lines():
                 yield json.loads(line)
+
+    def status(self) -> DatasetStatus:
+        """Retrieve this datast's streamability status.
+
+        :return: Dataset streamability status.
+        :rtype: DatasetStatus
```
pcattori:
```
:class:`~tamr_unify_client.models.dataset_status.DatasetStatus`
``` 
instead of just `DatasetStatus` so that docs render a link to the class.

tests/unit/test_dataset_by_external_id.py
```diff
@@ -0,0 +1,53 @@
+import json
+
+import pytest
+import responses
+
+from tamr_unify_client import Client
+from tamr_unify_client.auth import UsernamePasswordAuth
+
+_dataset_json = """[{
```
pcattori: If you want to avoid manually `dumps` and `loads` JSON, you could instead represent this as Python (i.e. `list`s and `dict`s) and use `responses.add(..., json=_dataset_json)` and then compare directly to `.data` (i.e. `actual_dataset.data == _dataset_json`).

tests/unit/test_project_by_external_id.py
```diff
@@ -0,0 +1,49 @@
...
+  "lastModified": {
+    "username": "admin",
+    "time": "2018-09-10T16:06:20.851Z",
+    "version": "project 1 modified version"
+  },
+  "relativeId": "projects/1"
+}]
+"""
+
+_project_external_id = "project 1 external ID"
```
pcattori: it's more pythonic to eliminate the leading `_` here since it's generally well understood that you shouldn't be importing/reusing stuff from test files.

tests/unit/test_dataset_by_external_id.py
```diff
@@ -0,0 +1,53 @@
...
+}]
+"""
+
+_dataset_external_id = "number 1"
+_datasets_url = f"http://localhost:9100/api/versioned/v1/datasets?filter=externalId=={_dataset_external_id}"
+
+
+@responses.activate
+def test_dataset_by_external_id__raises_when_not_found():
+    responses.add(responses.GET, _datasets_url, body="[]")
```
pcattori: `json=[]` instead of `body="[]"`

tamr_unify_client/models/base_collection.py
Andy-L: I picked `ValueError` because the error is that there are too many values, not that something was wrong with the key.  It's also something that shouldn't ever happen, because if it does it means the index in PG isn't enforcing uniqueness.

tamr_unify_client/models/dataset/resource.py
Andy-L: is that 
`:rtype: :class: ~tamr_unify_client.models.dataset_status.DatasetStatus`?

Andy-L: oic, it's
```
:rtype: :class:`tamr_unify_client.models.dataset_status.DatasetStatus`
```

tamr_unify_client/models/base_collection.py
Andy-L: fixed

tamr_unify_client/models/base_collection.py
Andy-L: added

tamr_unify_client/models/base_collection.py
Andy-L: let me know what you think; if I should change the second one to be `KeyError` (or something else), or if `ValueError` makes sense.

tamr_unify_client/models/dataset/resource.py
Andy-L: or do I need the `~` for something?

tests/unit/test_dataset_by_external_id.py
Andy-L: 👍

tests/unit/test_project_by_external_id.py
Andy-L: 🤷‍♂️ ok

tests/unit/test_dataset_by_external_id.py
Andy-L: fixed

tamr_unify_client/models/dataset/resource.py
pcattori: the `~` shortens the rendered link to just the last thing so `DatasetStatus` instead of `tamr_unify_client.models.dataset_status.DatasetStatus`. We've been using `~`/shortened links unless there's a good reason not to.

tamr_unify_client/models/base_collection.py
pcattori: TIL there's a built-in [`LookupError`](https://docs.python.org/3/library/exceptions.html#LookupError) in Python.

> The base class for the exceptions that are raised when a key or index used on a mapping or sequence is invalid: IndexError, KeyError. This can be raised directly by codecs.lookup().

I think we should use `LookupError` here

Andy-L: for both?

pcattori: No, `KeyError` for the 0 case is exactly what `KeyError` is designed for so I think being specific there is good. Just changing the `ValueError` to `LookupError` is what I propose.

tamr_unify_client/models/dataset/resource.py
Andy-L: so that would be:
```
:rtype: :class:~`tamr_unify_client.models.dataset_status.DatasetStatus`
```
?

pcattori: ```:rtype: :class:`~tamr_unify_client.models.dataset_status.DatasetStatus````. So `~` inside the ```````.

pcattori: You can install the docs requirements (`pip install -r doc/requirements`) and then run `make html` from within the `doc` dir and open up the resulting `_build/html/index.html` in your web browser if you want to see it render.

tamr_unify_client/models/base_collection.py
Andy-L: done

tamr_unify_client/models/dataset/resource.py
Andy-L: 👍

Andy-L:
I had to modify `developer-interface.rst` to get `DatasetStatus` to show up.  I also added headings for `Dataset` and `Project`, and alphabetized because they seemed to be in some random order.

Issue #33, opened 2019-02-12
Python version (and any other requirements) should be specified more visibly


nsylv:
Got the following error after running the quickstart example in the docs:
```
Traceback (most recent call last):
  File "test.py", line 1, in <module>
    from tamr_unify_client import Client
  File "/home/nikki/.local/lib/python2.7/site-packages/tamr_unify_client/__init__.py", line 1, in <module>
    from tamr_unify_client.client import Client
  File "/home/nikki/.local/lib/python2.7/site-packages/tamr_unify_client/client.py", line 69
    return f"{method} {url} : {response.status_code}"
                                                    ^
SyntaxError: invalid syntax
```

The issue is that I'm running python 2.7, but 3.6+ is required.  Maybe we could add a requirements section to the docs and github to help make this more visible?

pcattori:
Currently, the Python version is specified in 3 places:

1) [README](https://github.com/Datatamer/unify-client-python/blob/master/README.md) badge
2) [PyPI page](https://pypi.org/project/tamr-unify-client/) in the `Classifiers` section
3) [.travis.yml](https://github.com/Datatamer/unify-client-python/blob/master/.travis.yml) / [Travis CI tests](https://travis-ci.org/Datatamer/unify-client-python)

---

We could add a "Requirements" section to the [installation instructions](https://tamr-unify-python-client.readthedocs.io/en/stable/user-guide/installation.html) specifying the compatible Python versions to make it more visible.

pcattori:
#58 adds a Python compatibility message at the top of the installation docs.

#40 prevents `pip install tamr-unify-client` from installing `tamr-unify-client` on incompatible Python versions. NOTE: this only affects version 0.3.0 and newer. E.g. trying to `pip install tamr-unify-client` from Python2.7 will result in `tamr-unify-client` version 0.2.0 being installed.

PR #40, opened 2019-02-22
Read files with appropriate encodings

DerrickRice:
`pip3 install` failed for me due to this omission.

Also specify required python version >= 3.6 (related to #33 )

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,10 +2,10 @@ from setuptools import find_packages, setup
 import os
 
 project_root = os.path.dirname(__file__)
-with open(os.path.join(project_root, "VERSION.txt")) as f:
+with open(os.path.join(project_root, "VERSION.txt"), encoding="utf-8") as f:
     version = f.readline().rstrip()
 
-with open(os.path.join(project_root, "README.md"), "r") as f:
+with open(os.path.join(project_root, "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
 setup(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,10 +2,10 @@
 import os
 
 project_root = os.path.dirname(__file__)
-with open(os.path.join(project_root, "VERSION.txt")) as f:
+with open(os.path.join(project_root, "VERSION.txt"), "r", encoding="UTF-8") as f:
```
pcattori: nit: [Python Unicode docs](https://docs.python.org/3/howto/unicode.html) lowercase `"uft-8"`:

```suggestion
with open(os.path.join(project_root, "VERSION.txt"), "r", encoding="utf-8") as f:
```

pcattori: Nit: Could you actually remove the `"r"` from both since it's the default mode anyway?

DerrickRice:
@pcattori  - do you want these changes amended and push --force ? or stacked on top?

(Perhaps the [Contributor Guide](https://tamr-unify-python-client.readthedocs.io/en/stable/contributor-guide.html) can provide some guidance on your preferred practice)

pcattori:
@DerrickRice amended and `push -f` in this case. General stance is that we should keep the git history of the master branch clean when not too inconvenient to do so.

Definitely something we should add to the Contributor Guide!

DerrickRice:
ping @pcattori - changed per your feedback. Thanks!

PR #44, opened 2019-02-28
Fix #43: `request()` with an absolute path resolves incorrectly

### LINKED ISSUE #43: `request()` with an absolute path resolves incorrectly
When `request` is called with an absolute path (e.g. `/api/service/health`), it rewrites the path into a path relative to `base_path` instead of correctly resolving the absolute path.

Example:
`client.request(method, "/api/service/health")`

Desired result:
Request is made to `http://unify-host:9100/api/service/health`

Actual result:
Request is made to `http://unify-host:9100/api/versioned/v1//api/service/health`

nbateshaus:
Use `urllib.parse.urljoin()` to create URLs, so that absolute paths are handled correctly.

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -1,3 +1,4 @@
+from urllib.parse import urljoin
 import requests
 
 from tamr_unify_client.models.dataset.collection import DatasetCollection
@@ -51,7 +52,7 @@ class Client:
         host="localhost",
         protocol="http",
         port=9100,
-        base_path="api/versioned/v1",
+        base_path="api/versioned/v1/",
     ):
         self.auth = auth
         self.host = host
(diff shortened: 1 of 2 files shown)


VERSION.txt
```diff
@@ -1 +1 @@
-0.2.0
+0.3.0
```
pcattori: In our current release process, the version number only gets bumped during the release cycle, not during individual PRs (so we can batch multiple changes in 1 version bump). We are in the process of updating our release process, but for now, could you remove the corresponding commit and `push -f`?

tamr_unify_client/client.py
```diff
@@ -1,3 +1,5 @@
+from urllib.parse import urljoin
+
```
pcattori: nit: no newlines between imports from 3rd party libs please!

nbateshaus: IntelliJ / Pycharm REALLY wants them in there, but I will fix.

PR #50, opened 2019-03-05
Add flake8-import-order plugin

pcattori:
Running `flake .` in the root directory for this project now enforces [Google-style import ordering](https://google.github.io/styleguide/pyguide.html?showone=Imports_formatting#313-imports-formatting) ([example](https://github.com/PyCQA/flake8-import-order/blob/master/tests/test_cases/complete_google.py)).


pcattori:
@laferrieren black is specifically a "formatter that does not change the AST" but rearranging imports could change the AST. But yes, it would be nice to have a tool that does both...

PR #51, opened 2019-03-05
Updated release process

pcattori:


RELEASE.md
```diff
@@ -1,30 +1,55 @@
...
+
+Then, create a branch on Github within the [Datatamer/unify-client-python](https://github.com/Datatamer/unify-client-python) repo titled `release-<version>` e.g. `release-0.3.0`.
+
+NOTE: This release branch should *not* contain the version bump changes from Step 1.
+
+# 3. Remove `-dev` suffix on release branch
+
+Create a PR *to the release branch* with the following changes:
+- `VERSION.txt`: Remove `-dev` suffix from version e.g. `0.3.0-dev` -> `0.3.0`.
+- `CHANGELOG.md`: Remove the `-dev` suffix from the version being released e.g. `# 0.3.0-dev` -> `# 0.3.0`.
```
pcattori: @Andy-L I'm debating whether removing the `-dev` suffix in the changelog should be part of Step 3 or part of Step 1. Curious what you think.

RELEASE.md
```diff
@@ -1,30 +1,55 @@
...
-8. Merge release branch onto `master`
-9. Tag latest commit with version number on Github
-  `git tag -a v<version> -m "Release v<version>"`
-  `git push Datatamer --tags`
-10. Build release artifacts
-  `python setup.py sdist bdist_wheel`
-11. Publish to PyPI
-  `twine upload dist/*<version>*`
-    - If necessary, update the version/date of the corresponding entry in [CHANGELOG.md](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md)
+During the following steps we will releasing the `0.3.0` version as an example.
```
Andy-L: we will **be** releasing?

RELEASE.md
```diff
@@ -1,30 +1,55 @@
...
+
+For our example, that means `master` branch is currently on version `0.3.0-dev` (you can check `VERSION.txt` to see the version).
+
+Be sure to substitute `0.3.0` appropriately for the actual version being released.
```
Andy-L: `for` => `with`

RELEASE.md
Andy-L: I think it needs to be in both:

- on master, you want the changes for the now-prior version to be listed under it's released version, `0.3.0`
- similarly, on the release branch you also want the to-be-released changes to be listed under the released version, 0.3.0`

An alternative might be to never use the `-dev` suffix in the changelog itself.

RELEASE.md
```diff
@@ -1,30 +1,55 @@
...
+NOTE: This release branch should *not* contain the version bump changes from Step 1.
+
+# 3. Remove `-dev` suffix on release branch
+
+Create a PR *to the release branch* with the following changes:
+- `VERSION.txt`: Remove `-dev` suffix from version e.g. `0.3.0-dev` -> `0.3.0`.
+- `CHANGELOG.md`: Remove the `-dev` suffix from the version being released e.g. `# 0.3.0-dev` -> `# 0.3.0`.
+
+Merge your changes into the release branch e.g. `release-0.3.0`.
+
```
Andy-L: don't we want to mention checking the Travis test results before creating the release?

RELEASE.md
```diff
@@ -1,30 +1,55 @@
...
-  `python setup.py sdist bdist_wheel`
-11. Publish to PyPI
-  `twine upload dist/*<version>*`
-    - If necessary, update the version/date of the corresponding entry in [CHANGELOG.md](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md)
+During the following steps we will releasing the `0.3.0` version as an example.
+
+For our example, that means `master` branch is currently on version `0.3.0-dev` (you can check `VERSION.txt` to see the version).
+
+Be sure to substitute `0.3.0` appropriately for the actual version being released.
+
```
Andy-L: somewhere we should probably mention that the Travis tests should be all passing for the commit from which the release branch is branched.

RELEASE.md
pcattori: Good point. I went with specifying this in both Step 1 and Step 3.

PR #59, opened 2019-03-07
Improve repr of client object

DerrickRice:
# ↪️ Pull Request

Through example, I am proposing of a pattern for `__repr__` throughout the codebase.

This repr has strictly more information than the default object repr provided by Python, but makes no claims of being comprehensive or `eval`-able.

Related to #78 

## 💻 Examples

```python3
>>> from tamr_unify_client import Client
>>> c = Client(None)
>>> c
tamr_unify_client.client.Client(host='localhost', protocol='http', port=9100, base_path='api/versioned/v1/')@0x7fbc4a6f0780
```

## 🚨 Test instructions

unit tests added.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Update relevant docs + docstrings
- [x] Add ("Added", "FIXED", "REMOVED") entry/entries for the current development version in the changelog

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -140,3 +140,14 @@ class Client:
         :rtype: :class:`~tamr_unify_client.models.DatasetCollection`
         """
         return self._datasets
+
+    def __repr__(self):
+        # intentionally leaving out auth in order to avoid security concerns
+        return (
+            f"{self.__class__.__module__}."
+            f"{self.__class__.__qualname__}("
+            f"host={self.host!r}, "
+            f"protocol={self.protocol!r}, "
+            f"port={self.port!r}, "
+            f"base_path={self.base_path!r})"
+        )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.4.0-dev
   **ADDED**
   - Record Clusters API endpoint to finish working mastering workflow.
+  - [#78](https://github.com/Datatamer/unify-client-python/issues/78) Improved repr for Client instances
 
   **FIXED**
   - Mastering workflow example was missing the generate clusters step, which has been rectified using proper endpoint
--- a/tests/unit/test_strings.py
+++ b/tests/unit/test_strings.py
@@ -1,0 +1,21 @@
+from tamr_unify_client import Client
+from tamr_unify_client.auth import UsernamePasswordAuth
+
+
+def test_client_repr():
+    auth = UsernamePasswordAuth("username", "password")
+
+    unify = Client(auth)
+    rstr = f"{unify!r}"
+
+    assert rstr.startswith("tamr_unify_client.client.Client(")
+    assert "http" in rstr
+    assert rstr.endswith(")")
+    assert "password" not in rstr
+
+    unify = Client(auth, protocol="http", port=1234, base_path="foo/bar")
+    rstr = f"{unify!r}"
+
+    assert "'http'" in rstr
+    assert "1234" in rstr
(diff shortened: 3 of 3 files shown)


tamr_unify_client/client.py
```diff
@@ -140,3 +140,15 @@ def datasets(self):
         :rtype: :class:`~tamr_unify_client.models.DatasetCollection`
         """
         return self._datasets
+
+    def __repr__(self):
+        # intentionally leaving out auth in order to avoid security concerns
+        return (
+            f"{self.__class__.__module__}."
+            f"{self.__class__.__qualname__}("
```
pcattori: Thinking out loud:

We want users to import the client via `from tamr_unify_client import Client`. AFAIK `__qualname__` is intentionally designed to not be smart about re-finagling imports (e.g. [__init__.py](https://github.com/Datatamer/unify-client-python/blob/master/tamr_unify_client/__init__.py)).

That leads to the import being differently namespaced than the qualname.

One thing to consider: at some point we might want to re-finagle more imports in `__init__.py` to make importing easier (less unnecessary namespacing for users). If we do that, then getting rid of `__qualname__` here would make things consistent between the import and the repr.

---

Action: Let's keep this as is.

DerrickRice: `__qualname__` is only the class name, relative to it's module. In this case:
- `__qualname__` => Client
- `__module__` => tamr\_unify\_client.client

Certain classes could be manually "re-homed" if we want to show (for example) "tamr\_unify\_client" rather than "tamr\_unify\_client.client". This would need to be hard-coded.

> Action: Let's keep this as is.

That's the conclusion I reach as well. The simplest option that still provides consistency and clarity is to use the classes actual module, rather than re-homing.

Also, worth noting that the existing default `__repr__` of this object shows the full path as well. So it's at least not a departure from existing behavior. (Which leaves the door open for a different decision later, if desired)

DerrickRice:
@pcattori, if this looks like a viable strategy, I will put forward a comprehensive PR with the same approach applied to other objects. Then we can get into the details of what attrs to include or exclude on each class's `__repr__`.

tamr_unify_client/client.py
```diff
@@ -140,3 +140,15 @@ def datasets(self):
...
+            f"host={self.host!r}, "
+            f"protocol={self.protocol!r}, "
+            f"port={self.port!r}, "
+            f"base_path={self.base_path!r}"
+            f")@0x{id(self):x}"
```
pcattori: @DerrickRice Genuinely not sure how I feel about including the object ID in the `repr`. I can see how it could be needlessly confusing to beginners, but could also see how it could help distinguish different instances.

Do you know if showing the object ID in the `repr` is a common pattern?

DerrickRice: It is not a common pattern and I am not attached to it. It does trade brevity/simplicity for specificity.

I suspect that it's more likely that the overall cost of including it (i.e. confusion and verbosity) is higher than the risk of confusion when two objects have the same repr but aren't the same object.

It is common for `__repr__` to not distinguish between two exactly similar objects.

Let's remove it?

pcattori: 👍 Yes, I agree. Let's remove it.

pcattori:
> @pcattori, if this looks like a viable strategy, I will put forward a comprehensive PR with the same approach applied to other objects. Then we can get into the details of what attrs to include or exclude on each class's `__repr__`.

@DerrickRice sounds good! Let's get this one merged in first, but then I think what you propose makes total sense.

Issue #61, opened 2019-03-07
Change `data` to `_data`


DerrickRice:
# 💬 RFC

Many classes within tamr_unify_client use a `data` field member for the static information, often loaded from a server's JSON response upon fetching the object.

This object is not documented. Because of its direct relationship to the server response, it's also subject to change from version to version. It should be messaged as a private by the underscore naming convention (`_data`)

This also affects tab completion in the python3 interactive repl.

## 🔦 Context

This is particularly misleading when working with a dataset object, because it looks like it ought to be the collection of the dataset's records.

```python3
>>> dataset = project.unified_dataset()
>>> dataset.
dataset.api_path         dataset.data             dataset.external_id      dataset.from_json(       dataset.records(         dataset.relative_id      dataset.status(          dataset.update_records(
dataset.client           dataset.description      dataset.from_data(       dataset.name             dataset.refresh(         dataset.resource_id      dataset.tags             dataset.version
>>> dataset.
```

## 💻 Examples

- Change base_resource to use `_data`
- Change various `self.data` references to `self._data`

pcattori:
Go for launch! 🚀

PR #70, opened 2019-03-07
Rename data field to _data (private)

DerrickRice:
# ↪️ Pull Request

In response to RFC #61

`_data` contains information that is undocumented and likely to change without notice.

## 🚨 Test instructions

Existing unit tests updated. Searched, via grep, for all references to "data". The only remaining references are in the `__init__` param and corresponding docstr for BaseResource (as intended).

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Update relevant docs + docstrings
- [ ] Add ("Added", "FIXED", "REMOVED") entry/entries for the current development version in the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,14 @@
   **FIXED**
   - Mastering workflow example was missing the generate clusters step, which has been rectified using proper endpoint
   - [#30](https://github.com/Datatamer/unify-client-python/issues/30) Better docs for how to call directly call APIs
+  - [#61](https://github.com/Datatamer/unify-client-python/issues/61) `data` field renamed to `_data` (private).
+
+  **REMOVED**
+  - [#61](https://github.com/Datatamer/unify-client-python/issues/61) `data` field renamed to `_data` (private).
 
 ## 0.3.0
+*released on 2019-3-1*
+
   **ADDED**
   - Versioning example in FAQ
   - Offline installation docs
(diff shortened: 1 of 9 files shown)


DerrickRice:
Done. (Also noted the release date of 0.3.0 along the way)

CHANGELOG.md
```diff
@@ -4,8 +4,11 @@
 
   **FIXED**
   - Mastering workflow example was missing the generate clusters step, which has been rectified using proper endpoint
+  - `data` field renamed to `_data` (private). ([details](https://github.com/Datatamer/unify-client-python/issues/61))
```
pcattori: The removal of `.data` should be noted under a `**REMOVED**` header as well, since it will be a breaking change for those are currently using it.

nit: could you also rename your link from `details` to `PR`? We originally went with `details` in past changelog entries but that was too vague. sometimes we want to include the issue #, sometimes the PR #, and sometimes both.

pcattori: I see you copied the format some of the previous changelog entries.

I think we should move to a new format:
```suggestion
  - [#61](https://github.com/Datatamer/unify-client-python/issues/61) `data` field renamed to `_data` (private).
```

pcattori: ~We should probably add this format in the changelog item within PR template~ specified the changelog entry format in #73

Issue #78, opened 2019-03-13
Add useful repr to various object throughout the codebase


DerrickRice:
# 🙋 feature request

Desire: Add a useful repr() to all common objects throughout unify-client-python.

## 🤔 Expected Behavior

interactive python and `repr(x)` for objects from the library should show some elements that are useful for a developer, though they need not be eval-compatible.

## 😯 Current Behavior

Default Python `Object.__repr__`

## 💁 Possible Solution

Discussed and negotiated in PR #59

DerrickRice:
Landing changes in PR #59 for Client, then proposing a bigger PR for other objects.

Later PR will result in a broader scope for the Changelog.

pcattori:
Implemented in #59 and #83

Issue #62, opened 2019-03-07
MasteringProject high_impact_pairs and published_clusters not loaded on fetch


DerrickRice:
# 🐛 bug report

`MasteringProject.high_impact_pairs` and `MasteringProject.published_clusters` do not fetch the dataset information from the server.

This is inconsistent with `Project.unified_dataset`

## 😯 Current Behavior

```python3
>>> project.high_impact_pairs().external_id
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/drice/c/tamr/unify-client-python/tamr_unify_client/models/dataset/resource.py", line 23, in external_id
    return self.data["externalId"]
TypeError: 'NoneType' object is not subscriptable

>>> project.published_clusters().external_id
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/drice/c/tamr/unify-client-python/tamr_unify_client/models/dataset/resource.py", line 23, in external_id
    return self.data["externalId"]
TypeError: 'NoneType' object is not subscriptable

>>> project.unified_dataset().external_id
'idogs_unified_dataset'
```

## 🤔 Expected Behavior / 💁 Possible Solution

use the same logic as `Project.unified_dataset`. Fetch the dataset object from the server, so that the returned `Dataset` has all of the appropriate fields and does not generate exceptions on standard property access.

## 🌍 Your Environment

| Software          | Version(s) |
| ----------------- | ---------- |
| Python            | 3.6
| tamr-unify-client | 0.4.0-dev (master HEAD at the time of writing)
| Operating System  | Linux (ubuntu:latest docker container)

PR #83, opened 2019-03-14
Useful reprs for all objects.

DerrickRice:
# ↪️ Pull Request

Issue #78 

The goal is to show useful values for developers during exploration,
debugging, or logging. These reprs do not show all possible
distinguishing values. The repr format is not documented and may change
without notice.

## 💻 Examples

```text
unify: tamr_unify_client.client.Client(host='localhost', protocol='https', port=4443, base_path='api/versioned/v1/', auth=UsernamePasswordAuth)
project: tamr_unify_client.models.project.mastering.MasteringProject(relative_id='projects/3', name='idogs', type='DEDUP')
```

## ✔️ PR Todo

- [X] Added/updated testing for this change
- [X] Update relevant [docs](https://github.com/Datatamer/unify-client-python/tree/master/docs) + docstrings
- [X] Update the [CHANGELOG](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md) under the current `-dev` version:
  - Add changelog entries under any that apply: **ADDED**, **FIXED**, **REMOVED**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.4.0-dev
   **BREAKING CHANGES**
   - [#61](https://github.com/Datatamer/unify-client-python/issues/61) `data` field renamed to `_data` (private).
+  - [#78](https://github.com/Datatamer/unify-client-python/issues/78) Property accessors return `None` rather than raise `KeyError`
 
   **NEW FEATURES**
   - Record Clusters API endpoint to finish working mastering workflow.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
   **NEW FEATURES**
   - Record Clusters API endpoint to finish working mastering workflow.
-  - [#78](https://github.com/Datatamer/unify-client-python/issues/78) Improved repr for Client instances
+  - [#78](https://github.com/Datatamer/unify-client-python/issues/78) Improved repr for objects through the library
   - [#42](https://github.com/Datatamer/unify-client-python/issues/42) Optional `session` argument to `Client` to use a specific `requests.Session` instance
 
   **BUG FIXES**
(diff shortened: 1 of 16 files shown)


DerrickRice:
Currently, some repr's can raise exceptions. reprs should endeavor to be exception-free.

The cause of the exception is missing `data` values. (See #62 )

Possible paths forward:

- Catch certain exceptions (i.e. KeyError) in `__repr__` methods
- change property getter to use `self.data.get('foo')` which will return `None`, rather than fail-fast on missing values.
- change `__repr__` to use `self.data.get('foo')` rather than `self.foo` property getter.

I suggest either the second or third oprtion. The disadvantage with the first is that it hides all other useful information if just one getter fails.

I propose tacking the relevant change onto this pull request as a separate commit. Waiting for comment from the maintainers.

pcattori:
@DerrickRice , agreed. Let's go with approach 2 (changing property getters to use `.get`)

DerrickRice:
@pcattori - exception-free changes tacked on. As an internal change, I didn't add a changelog entry for that particular commit.

pcattori:
@DerrickRice there's a slight breaking change here in that `KeyError` is no longer thrown by property getters, but rather `None` can be returned instead now. Could you add a message to that effect under **REMOVED** in the changelog?

---

P.S. Realizing that **REMOVED** might not be the best name for the section. Instead, we could say:

1. **BREAKING CHANGES**, **NEW FEATURES**, **BUG FIXES**
2. **CHANGED**, **ADDED**, **FIXED**

I like (1) best.

DerrickRice:
> BREAKING CHANGES, NEW FEATURES, BUG FIXES

Sounds good.

I can imagine some things being debatable between bug fixes and breaking changes ("Is the new behavior the one that was always intended, therefore it's a bug fix? Or is the new behavior one that is a recent decision, therefore it's a breaking change?")

I expect most things to fall into a single category, though.

pcattori:
@DerrickRice just merged #85 so that the new changelog sections naming/ordering is consistent with PR templates. Could you rebase and order the sections as described in #85?

DerrickRice:
I think we're good to go, @pcattori

Issue #64, opened 2019-03-07
Conflation between 'api_path', 'relative_id' / 'relativeId', and BaseResource ctor 'alias'


DerrickRice:
# 💬 RFC

These terms are used nearly interchangeably:
- `Object.relative_id` (which reads `data["relativeId"]`)
- `Object.api_path`
- argument `alias` on `BaseResource.__init__`

## 🔦 Context

"relative ID" is an unfamiliar term.

I understand relative path, and api path. Those describe things that I'm familiar with.

If I had to vote for one, I'd request `api_path`, since it is tacked onto the end of the versioned API path.

Further confusing: sometimes `resource.api_path` is valid and populated, and sometimes different from `resource.relative_id`. This has to do with the path the resource was fetched from vs the canonical path that the resource lives at. For example, unified data sets.

```python3
>>> ds = project.unified_dataset()
>>> ds.api_path
'projects/3/unifiedDataset'
>>> ds.relative_id
'datasets/58'
```

How is this meant to be used?

## 💻 Examples

Before:
```python3
>>> client.datasets.by_relative_id('datasets/58')
<tamr_unify_client.models.dataset.resource.Dataset object at 0x7fe17cddee80>
```

After:
```python3
>>> client.datasets.by_api_path('datasets/58')
<tamr_unify_client.models.dataset.resource.Dataset object at 0x7fe17cddee80>
```

pcattori:
Example:

`/api/versioned/v1/datasets/4` might be the canonical path for a dataset but if it happens to also be the unified dataset for project 1, then `/api/versioned/v1/projects/1/unifiedDataset` is an alias for that same resource.

Note that we wouldn't have to treat these cases specially if the API did the same things no matter which way you refer to the resource, but for our API if you do `/api/versioned/v1/datasets/4:refresh` it does a different thing than `/api/versioned/v1/projects/1/unifiedDataset:refresh`.

Keeping a reference to the (canonical) `relative_id` as well as the `alias` allows both behaviors to be possible.

PR #95, opened 2019-04-25
Access to Attributes of a Dataset

### LINKED ISSUE #94: Access to Attributes of a Dataset
# 🙋 feature request

Attribute resources, available individually or as a collection under Dataset.
https://docs.tamr.com/reference#attribute-types

## 🤔 Expected Behavior

Access attributes collection as a sub-resource of any dataset.
Access an attribute as a resource.

## 😯 Current Behavior

No way to get attributes.

## 💁 Possible Solution

Implement it.

## 🔦 Context

I need to know attribute types in order to do intelligent conversion between Unify values and values in the outside world.

In particular, I need to identify attributes that use Unify's geometry representation so I can convert between those and the Python Geo Interface.

## 💻 Examples

```
dataset = client.datasets().by_name("my_dataset")
for attr in dataset.attributes():
    do_something_with(attr)

attr = client.datasets().by_name("my_dataset").attributes().by_name("my_attribute")
print(attr.name)
print(attr.description)
print(attr.nullable)
```
Should produce something like:
```
"my_attribute"
"Description of my_attribute"
True
```

nbateshaus:
# ↪️ Pull Request

Access to Attributes of a Dataset.
Fix #94 

## 💻 Examples

```
>>> dataset1 = unify.datasets.by_resource_id("1")
>>> for attribute in dataset1.attributes:
...   print(attribute.relative_id)
...   print(attribute.name)
...   print(attribute.type.base_type)
... 
datasets/1/attributes/RowNum
RowNum
STRING
datasets/1/attributes/geo
geo
RECORD
```

## ✔️ PR Todo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## 0.5.0-dev
+  **NEW FEATURES**
+  - [#94](https://github.com/Datatamer/unify-client-python/issues/94) Add access to Attributes of a Dataset
 
 ## 0.4.0
   **BREAKING CHANGES**
--- a/docs/developer-interface.rst
+++ b/docs/developer-interface.rst
@@ -31,6 +31,21 @@ Datasets
 .. autoclass:: tamr_unify_client.models.dataset.collection.DatasetCollection
   :members:
 
+Attribute
+---------
+
+.. autoclass:: tamr_unify_client.models.attribute.resource.Attribute
+
+Attribute Type
+--------------
+
+.. autoclass:: tamr_unify_client.models.attribute.type.AttributeType
+
+Attributes
+----------
+
+.. autoclass:: tamr_unify_client.models.attribute.collection.AttributeCollection
+
 Machine Learning Models
 -----------------------
 
--- a/tamr_unify_client/models/attribute/collection.py
+++ b/tamr_unify_client/models/attribute/collection.py
@@ -43,7 +43,9 @@ class AttributeCollection(BaseCollection):
         """
         split_id = relative_id.split("/")
         if split_id[:-1] != self.api_path:
-            raise ValueError(f"Attribute f{relative_id} is not in collection {self.api_path}")
+            raise ValueError(
+                f"Attribute f{relative_id} is not in collection {self.api_path}"
+            )
         resource_id = split_id[-1]
         return self.by_resource_id(resource_id)
 
@@ -63,17 +65,17 @@ class AttributeCollection(BaseCollection):
         raise NotImplementedError("Attributes do not have external_id")
 
     def stream(self):
-        """Stream datasets in this collection. Implicitly called when iterating
+        """Stream attributes in this collection. Implicitly called when iterating
         over this collection.
 
-        :returns: Stream of datasets.
-        :rtype: Python generator yielding :class:`~tamr_unify_client.models.dataset.resource.Dataset`
+        :returns: Stream of attributes.
+        :rtype: Python generator yielding :class:`~tamr_unify_client.models.attribute.resource.Attribute`
 
         Usage:
-            >>> for dataset in collection.stream(): # explicit
-            >>>     do_stuff(dataset)
-            >>> for dataset in collection: # implicit
-            >>>     do_stuff(dataset)
+            >>> for attribute in collection.stream(): # explicit
+            >>>     do_stuff(attribute)
+            >>> for attribute in collection: # implicit
+            >>>     do_stuff(attribute)
         """
         for resource_json in self._data:
             alias = self.api_path + "/" + resource_json["name"]
--- a/tamr_unify_client/models/attribute/collection.py
+++ b/tamr_unify_client/models/attribute/collection.py
@@ -41,12 +41,7 @@ class AttributeCollection(BaseCollection):
         :returns: The specified attribute.
         :rtype: :class:`~tamr_unify_client.models.attribute.resource.Attribute`
         """
-        split_id = relative_id.split("/")
-        if split_id[:-1] != self.api_path:
-            raise ValueError(
-                f"Attribute f{relative_id} is not in collection {self.api_path}"
-            )
-        resource_id = split_id[-1]
+        resource_id = relative_id.split("/")[-1]
         return self.by_resource_id(resource_id)
 
     def by_external_id(self, external_id):
--- a/tamr_unify_client/models/attribute/collection.py
+++ b/tamr_unify_client/models/attribute/collection.py
@@ -7,7 +7,8 @@ class AttributeCollection(BaseCollection):
 
     :param client: Client for API call delegation.
     :type client: :class:`~tamr_unify_client.Client`
-    :param :py:class:`dict` data: JSON data representing this resource
+    :param data: JSON data representing this resource
+    :type data: dict
     :param api_path: API path used to access this collection.
         E.g. ``"datasets/1/attributes"``.
     :type api_path: str
@@ -48,7 +49,7 @@ class AttributeCollection(BaseCollection):
         """Retrieve an attribute by external ID.
 
         Since attributes do not have external IDs, this method is not supported and will
-        raise a NotImplementedError.
+        raise a :class:`NotImplementedError` .
 
         :param external_id: The external ID.
         :type external_id: str
--- a/tamr_unify_client/models/attribute/collection.py
+++ b/tamr_unify_client/models/attribute/collection.py
@@ -1,0 +1,96 @@
+from tamr_unify_client.models.attribute.resource import Attribute
+from tamr_unify_client.models.base_collection import BaseCollection
+
+
+class AttributeCollection(BaseCollection):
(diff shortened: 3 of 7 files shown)


tamr_unify_client/models/attribute/collection.py
```diff
@@ -0,0 +1,96 @@
...
+    def by_relative_id(self, relative_id):
+        """Retrieve an attribute by relative ID.
+
+        :param relative_id: The resource ID. E.g. ``"datasets/1/attributes/AttributeName"``
+        :type relative_id: str
+        :returns: The specified attribute.
+        :rtype: :class:`~tamr_unify_client.models.attribute.resource.Attribute`
+        """
+        split_id = relative_id.split("/")
+        if split_id[:-1] != self.api_path:
```
nbateshaus: This might be too aggressive.
For example, `datasets/1/attributes` might be the same as `projects/1/unifiedDataset/attributes`.

tamr_unify_client/models/attribute/collection.py
```diff
@@ -0,0 +1,96 @@
...
+        raise a NotImplementedError.
+
+        :param external_id: The external ID.
+        :type external_id: str
+        :returns: The specified attribute, if found.
+        :rtype: :class:`~tamr_unify_client.models.attribute.resource.Attribute`
+        :raises KeyError: If no attribute with the specified external_id is found
+        :raises LookupError: If multiple attributes with the specified external_id are found
+        """
+        raise NotImplementedError("Attributes do not have external_id")
```
nbateshaus: I looked at KeyError, TypeError, and setting `by_external_id=None`; none of those options seemed better than NotImplementedError.

tamr_unify_client/models/attribute/collection.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def stream(self):
+        """Stream datasets in this collection. Implicitly called when iterating
```
nbateshaus: Error in comment: `datasets` should be `attributes`.

tamr_unify_client/models/attribute/resource.py
```diff
@@ -0,0 +1,48 @@
...
+
+class Attribute(BaseResource):
+    """
+    A Unify Attribute.
+
+    See https://docs.tamr.com/reference#attribute-types
+    """
+
+    @classmethod
+    def from_json(cls, client, data, api_path):
```
nbateshaus: Note: unlike other resources, `api_path` is required, because there is not default collection of attributes.

tamr_unify_client/models/attribute/resource.py
```diff
@@ -0,0 +1,48 @@
...
+        return super().from_data(client, data, api_path)
+
+    @property
+    def relative_id(self):
+        return self.api_path
```
nbateshaus: The contract around `api_path` is unclear, so it's unclear whether this equivalence between `api_path` and `relative_id` is valid.

tamr_unify_client/models/attribute/type.py
```diff
@@ -0,0 +1,45 @@
...
+    @property
+    def base_type(self):
+        """:type: str"""
+        return self._data.get("baseType")
+
+    @property
+    def inner_type(self):
+        """:type: :class:`~tamr_unify_client.models.attribute.type.AttributeType`"""
+        if "innerType" in self._data:
+            alias = self.api_path + "/type"
```
nbateshaus: Author judgment: inserting `type` here gives a distinct URI to the type, but ends up with URIs like:
`datasets/1/attributes/id/geom/type/attributes/point/type`
I believe this is correct.

tamr_unify_client/models/attribute/type.py
```diff
@@ -0,0 +1,45 @@
...
+        # Import locally to avoid circular dependency
+        from tamr_unify_client.models.attribute.collection import AttributeCollection
+        return AttributeCollection.from_json(self.client, collection_json, alias)
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__module__}."
+            f"{self.__class__.__qualname__}("
+            f"relative_id={self.relative_id!r}, "
+            f"base_type={self.base_type!r})"
```
nbateshaus: Author judgment: I left inner_type and attributes out of the __repr__.

docs/developer-interface.rst
```diff
@@ -31,6 +31,21 @@ Datasets
...
+Attribute Type
+--------------
+
+.. autoclass:: tamr_unify_client.models.attribute.type.AttributeType
+
+Attributes
+----------
+
+.. autoclass:: tamr_unify_client.models.attribute.collection.AttributeCollection
+
```
Lordluen: LGTM

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 ## 0.5.0-dev
+  **NEW FEATURES**
+  - [#94](https://github.com/Datatamer/unify-client-python/issues/94) Add access to Attributes of a Dataset
```
Lordluen: LGTM

tamr_unify_client/models/attribute/collection.py
```diff
@@ -0,0 +1,94 @@
...
+
+class AttributeCollection(BaseCollection):
+    """Collection of :class:`~tamr_unify_client.models.attribute.resource.Attribute` s.
+
+    :param client: Client for API call delegation.
+    :type client: :class:`~tamr_unify_client.Client`
+    :param data: JSON data representing this resource
+    :type data: dict
+    :param api_path: API path used to access this collection.
+        E.g. ``"datasets/1/attributes"``.
```
Lordluen: I was going to comment that a comma should follow E.g., but according to [this](https://www.dailywritingtips.com/comma-after-i-e-and-e-g/) article, it is not an agreed upon rule. So LGTM.

tamr_unify_client/models/attribute/collection.py
```diff
@@ -0,0 +1,94 @@
+from tamr_unify_client.models.attribute.resource import Attribute
+from tamr_unify_client.models.base_collection import BaseCollection
+
+
+class AttributeCollection(BaseCollection):
+    """Collection of :class:`~tamr_unify_client.models.attribute.resource.Attribute` s.
+
+    :param client: Client for API call delegation.
+    :type client: :class:`~tamr_unify_client.Client`
+    :param data: JSON data representing this resource
```
Lordluen: Given that this is specifically an attribute resource, I feel like saying just "resource" is a little vague. Instead, I would prefer "attribute resource". However, I know that is being a little nit-picky.

tamr_unify_client/models/attribute/resource.py
pcattori: `api_path` is `alias` when it exists, and `relative_id` when it does not. this distinction is useful for things like refreshing a unified dataset, where using the `relative_id` would hit `/datasets/{id}:refresh` rather than `/projects/{id}/unifiedDataset:refresh`. Agree that it's confusing. There's an existing RFC for suggestions to improve this #64

tamr_unify_client/models/attribute/collection.py
Lordluen: LGTM

tamr_unify_client/models/attribute/collection.py
```diff
@@ -0,0 +1,94 @@
...
+        over this collection.
+
+        :returns: Stream of attributes.
+        :rtype: Python generator yielding :class:`~tamr_unify_client.models.attribute.resource.Attribute`
+
+        Usage:
+            >>> for attribute in collection.stream(): # explicit
+            >>>     do_stuff(attribute)
+            >>> for attribute in collection: # implicit
+            >>>     do_stuff(attribute)
```
Lordluen: Nice documentation!

tamr_unify_client/models/attribute/resource.py
Lordluen: Should there be? That's a question for other people, so LGTM.

tamr_unify_client/models/attribute/resource.py
Lordluen: I think that is temporarily true. Not sure this will last forever though. Might be worth an in-code `# NOTE: For now api_path and relative_id are the same, but this may change in the future.` to help someone fixing a future bug.

tamr_unify_client/models/attribute/resource.py
nbateshaus: Nope, attributes exist only in context. That context might be:
* A top-level dataset
* The type of another attribute
Attributes in different contexts can have the same local id (name) but different types etc.

tamr_unify_client/models/attribute/collection.py
nbateshaus: I just cribbed it from DatasetCollection, so I can't take credit.

tamr_unify_client/models/attribute/resource.py
nbateshaus: Done.

PR #96, opened 2019-04-26
Implement Python Geo Interface on Dataset

nbateshaus:
# ↪️ Pull Request

Provide Python Geo Interface on Dataset
Fix #98

## 💻 Examples

To produce a GeoJSON representation of a dataset::
```
  dataset = client.datasets.by_name("my_dataset")
  with open("my_dataset.json", "w") as f:
    json.dump(dataset.__geo_interface__, f)
```
To stream the records of a dataset as Geospatial features::
```
  dataset = client.datasets.by_name("my_dataset")
  for feature in dataset.iterfeatures():
    do_something(feature)
```
And this is super spiffy:
```
>>> import geopandas
>>> df=geopandas.GeoDataFrame.from_features(unify.datasets.by_name('my_dataset'))
>>> df.head()
  BASE_ELEV ELEV_GL ELEV_SL TOP_GL TOP_SL    TYPE                                           geometry last_edited_date last_edited_user
0      38.8    55.1      94   62.2  101.1  BORDER  POLYGON ((-71.1522084419519 42.3745215835176, ...       2018-03-26           JAMERO
1      43.5    25.5      69   37.3   80.9  BORDER  POLYGON ((-71.1514340605336 42.3744385428379, ...       2018-03-26           JAMERO
2      41.8    23.9    65.7   37.4   79.2  BORDER  POLYGON ((-71.1516949801055 42.3743460611468, ...       2018-03-26           JAMERO
3      39.9    22.5    62.4     39   78.9  BORDER  POLYGON ((-71.15185254313759 42.3743409491551,...       2018-03-26           JAMERO
4      45.7    25.7    71.4   37.9   83.6  BORDER  POLYGON ((-71.1512542097983 42.3745055126458, ...       2018-03-26           JAMERO
```
## ✔️ PR Todo

- [x] Added/updated testing for this change
- [x] Included links to related issues/PRs
- [x] Update relevant [docs](https://github.com/Datatamer/unify-client-python/tree/master/docs) + docstrings
- [x] Update the [CHANGELOG](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md) under the current `-dev` version:

PR #99, opened 2019-04-30
New Feature: Update a Dataset from geospatial features

### LINKED ISSUE #100: Dataset should be updatable from Python Geo FeatureCollection
# 🙋 feature request

Since a Dataset can produce a Python Geo FeatureCollection, it should also be updatable from a Python Geo FeatureCollection.
See https://gist.github.com/sgillies/2217756

Relates to #98 

## 🤔 Expected Behavior

When given a FeatureCollection, Dataset should upsert records by matching the feature `id` to the record key.

## 😯 Current Behavior

No support.

## 💁 Possible Solution

Provide a `from_geo_features(...)` method on Dataset.

## 🔦 Context

When working with geospatial data, most Python GIS tools are able to produce a FeatureCollection. Dataset should embody the One True Way to convert from a FeatureCollection to a Dataset.

## 💻 Examples

```
>>> import geopandas
>>> my_dataset = unify.datasets.by_name("my_dataset")
>>> my_geo_data_frame = geopandas.GeoDataFrame.from_features(my_dataset)
>>> # Modify my_geo_data_frame
>>> my_dataset.from_geo_features(my_geo_data_frame)
```

nbateshaus:
# ↪️ Pull Request

Update a dataset from geospatial features
Depends on #96 
Fix #100

## 💻 Examples

```
>>> import geopandas
>>> my_dataset = unify.datasets.by_name('my_dataset')
>>> df=geopandas.GeoDataFrame.from_features(my_dataset)
>>> # Modify df
>>> my_dataset.from_geo_features(df)
```

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/models/dataset/resource.py
+++ b/tamr_unify_client/models/dataset/resource.py
@@ -94,6 +94,44 @@ class Dataset(BaseResource):
             self.client, status_json, api_path=self.api_path + "/status"
         )
 
+    def from_geo_features(self, features):
+        """Upsert this dataset from a geospatial FeatureCollection or iterable of Features.
+
+        `features` can be:
+
+        - An object that implements ``__geo_interface__`` as a FeatureCollection
+          (see https://gist.github.com/sgillies/2217756)
+        - An iterable of features, where each element is a feature dictionary or an object
+          that implements the ``__geo_interface__`` as a Feature
+        - A map where the "features" key contains an iterable of features
+
+        See: geopandas.GeoDataFrame.from_features()
+
+        :param features: geospatial features
+        """
+        if hasattr(features, "__geo_interface__"):
+            features = features.__geo_interface__
+        if hasattr(features, "get") and features.get("type") == "FeatureCollection":
+            features = features["features"]
+
+        key_attrs = self.key_attribute_names
+        if len(key_attrs) == 1:
+            record_id = "recordId"
+        else:
+            record_id = "compositeRecordId"
+
+        def upsert_stream(features):
+            for feature in features:
+                yield {
+                    "action": "CREATE",
+                    record_id: feature["id"],
+                    "record": self._feature_to_record(
+                        feature, key_attrs, self._geo_attr
+                    ),
+                }
+
+        self.update_records(upsert_stream(features))
+
     @property
     def __geo_interface__(self):
         """Retrieve a representation of this dataset that conforms to the Python Geo Interface.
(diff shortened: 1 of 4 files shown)


tamr_unify_client/models/dataset/resource.py
```diff
@@ -94,6 +94,188 @@ def status(self) -> DatasetStatus:
...
+        """A generator of the records in this dataset represented as geospatial features
+
+        See https://gist.github.com/sgillies/2217756
+
+        :return: stream of features
+        :rtype: Python generator yielding :py:class:`dict[str, object]`
+        """
+        key_attrs = self.key_attribute_names
+        if len(key_attrs) == 1:
+
```
Lordluen: style nit: so much whitespace!!!

tamr_unify_client/models/dataset/resource.py
```diff
@@ -94,6 +94,188 @@ def status(self) -> DatasetStatus:
...
+        :param record: Unify record
+        :param key_value: Function to extract the value of the primary key from the record
+        :param key_attrs: Set of attributes that comprise the primary key for the record
+        :param geo_attr: The singular attribute to use as the geometry
+        :return: map from str to object
+        """
+        feature = {"type": "Feature", "id": key_value(record)}
+        reserved = {"bbox", geo_attr}.union(key_attrs)
+        conversion = {
```
Lordluen: I feel like this set of possible geometries is repeated a lot (in PR #96 also). Maybe we could make that are more global object, and then you either choose to only use the keys or keys/values where necessary to avoid repetition?

tamr_unify_client/models/dataset/resource.py
```diff
@@ -94,6 +94,188 @@ def status(self) -> DatasetStatus:
...
+            feature = feature.__geo_interface__
+
+        record = {}
+
+        props = feature.get("properties")
+        if props:
+            for prop in props:
+                record[prop] = props[prop]
+
+        geometry = feature.get("geometry")
```
Lordluen: Can a feature exist without a geometry and/or bbox? I feel like there should be an assert somewhere to confirm this is a valid geometry.

nbateshaus: Sadly, both geometry and bbox are optional in a feature.
A FeatureCollection with neither is an ordinary dataset.

tamr_unify_client/models/dataset/resource.py
nbateshaus: Yep, now that #96 is merged I will rebase and fix this.

tamr_unify_client/models/dataset/resource.py
nbateshaus: Black style requires all this whitespace.

nbateshaus: I tried to do it as a lambda, but it was enough harder to read that I went with in-line function definition instead.

Lordluen:
LGTM

Issue #90, opened 2019-04-10
dataset.update_records does not return a reponse


keziah-tamr:
# 🐛 bug report

calling my_dataset.update_records(stuff) returns None

## 🤔 Expected Behavior

calling my_dataset.update_records(stuff) should return the result of "self.client.post(self.api_path + ":updateRecords", data=body)"

## 🌍 Your Environment

| Software          | Version(s) |
| ----------------- | ---------- |
| tamr-unify-client |
| Tamr Unify server |
| Python            |
| Operating System  |

PR #103, opened 2019-05-06
fix update_records

### LINKED ISSUE #101: Dataset.update_records should stream updates
# 🙋 feature request

At the moment `Dataset.update_records()` does a
```
  body = "\n".join([json.dumps(r) for r in records])
```
which materializes all the updates as one massive string.

## 🤔 Expected Behavior

Updates from a streaming source, e.g. a database, don't need to be materialized en-route to Unify.

## 😯 Current Behavior

Updates from a streaming source, e.g. a database, get materialized as one giant string en-route to Unify.

## 💁 Possible Solution

`requests` is able to stream from a generator, so this could be changed to
```
    def _stringify_updates(updates):
        for update in updates:
            yield f"{update}\n".encode("utf-8")

    self.client.post(
        self.api_path + ":updateRecords",
        headers={"Content-Encoding": "utf-8"},
        data=_stringify_updates(records)
    )
```

## 🔦 Context

Materializing records doesn't scale, and is slow.
I'm reading records from a database (this is simplified):
```
    def load_data_from_database():
        engine = sqlalchemy.create_engine(database.get_uri())
        with closing(engine.connect().execution_options(stream_results=True)) as conn:
                query = sqlalchemy.sql.text(sql)
                query = query.bindparams(**sql_params)
                with closing(conn.execute(query)) as cursor:
                    my_dataset.update_records(cursor)
```
Note that `stream_results=True` - this means the cursor will return one (really a few) record at a time so I don't use crazy amounts of RAM. This is defeated by python-client materializing everything.

## 💻 Examples

See above.

mollysacks:
# ↪️ Pull Request

Resolves #101 and #90 

## 💻 Examples

Allows clients to use update_records method without materializing the streaming source/ database into one large string. 

## ✔️ PR Todo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.5.0-dev
   **NEW FEATURES**
   - [#94](https://github.com/Datatamer/unify-client-python/issues/94) Add access to Attributes of a Dataset
-  - [#103](https://github.com/Datatamer/unify-client-python/issues/103) Dataset `update_records` improvements
+  - [#103](https://github.com/Datatamer/unify-client-python/issues/103) Dataset `update_records` now returns the JSON response body for the underlying `POST datasets/{id}:updateRecords` call
   - [#98](https://github.com/Datatamer/unify-client-python/issues/98) Add `__geo_interface__` to Dataset
   - [#100](https://github.com/Datatamer/unify-client-python/issues/100) Add `from_geo_features` to Dataset
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.5.0-dev
   **NEW FEATURES**
   - [#94](https://github.com/Datatamer/unify-client-python/issues/94) Add access to Attributes of a Dataset
+  - [#103](https://github.com/Datatamer/unify-client-python/issues/103) Dataset `update_records` improvements
 
 ## 0.4.0
   **BREAKING CHANGES**
--- a/tamr_unify_client/models/dataset/resource.py
+++ b/tamr_unify_client/models/dataset/resource.py
@@ -62,14 +62,20 @@ class Dataset(BaseResource):
         :returns: JSON response body from server.
         :rtype: :py:class:`dict`
         """
+
         def _stringify_updates(updates):
             for update in updates:
-                yield f'{update}'.encode('utf-8')
+                yield f"{update}".encode("utf-8")
 
-        return self.client.post(
-            self.api_path + ':updateRecords',
-            headers={'Content-Encoding': "utf-8"},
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 0.5.0-dev
   **NEW FEATURES**
   - [#94](https://github.com/Datatamer/unify-client-python/issues/94) Add access to Attributes of a Dataset
+  - [#103](https://github.com/Datatamer/unify-client-python/issues/103) Dataset `update_records` improvements
```
pcattori: Can you add more detail here? Perhaps something like "`update_records` now returns the JSON response body for the underlying `POST datasets/{id}:updateRecords` call"

tamr_unify_client/models/dataset/resource.py
```diff
@@ -60,8 +60,14 @@ def update_records(self, records):
         :param records: Each record should be formatted as specified in the `Public Docs for Dataset updates <https://docs.tamr.com/reference#modify-a-datasets-records>`_.
         :type records: iterable[dict]
```
pcattori: You should include a `:return:` and `:rtype:` in the docstring for `update_records` to reflect the new behavior. There are plenty of examples of this in the code base e.g. the `attributes` function just above `update_records`.

tests/unit/test_dataset_geo.py
```diff
@@ -537,8 +537,9 @@ def update_callback(request, snoop):
                 "record": {"geom": {"point": [1, 1]}, "id1": "2", "id2": "b"},
             },
         ]
-        expected = "\n".join(json.dumps(update) for update in updates)
-        actual = snoop["payload"]
+        expected = [json.dumps(update).replace('"', "'") for update in updates]
+        self.maxDiff = None
```
pcattori: @mollysacks what does setting `maxDiff` here do?

tests/unit/test_dataset_geo.py
```diff
@@ -537,8 +537,9 @@ def update_callback(request, snoop):
...
+        actual = [a.decode("utf8") for a in list(snoop["payload"])]
```
pcattori: no need to use `list` if you are immediately going to use a list comprehension:

```suggestion
        actual = [a.decode("utf8") for a in snoop["payload"]]
```

This also goes for the other occurrence of `snoop["payload"]`

tests/unit/test_dataset_geo.py
mollysacks: that was so I could look at the entire list, I forgot to delete it but I can go back and do that

pcattori: yes, please do delete 😄

Issue #106, opened 2019-05-22
Add support for initializing a source dataset


mdonovan-tamr:
# 🙋 feature request

PR #137, opened 2019-06-13
Ability to initialize source dataset

mollysacks:
# ↪️ Pull Request
Addresses #106 

## ✔️ PR Todo

### DIFF
--- a/tests/unit/test_create_dataset.py
+++ b/tests/unit/test_create_dataset.py
@@ -16,7 +16,7 @@ def test_create_dataset():
         "name": "dataset",
         "keyAttributeNames": ["F1"],
         "description": "So much data in here!",
-        "externalId": "Dataset created with pubapi",
+  			"externalId": "Dataset created with pubapi"
     }
 
     datasets_url = f"http://localhost:9100/api/versioned/v1/datasets"
@@ -27,4 +27,7 @@ def test_create_dataset():
 
     u = unify.create_dataset(dataset_config)
     p = unify.datasets.by_resource_id("1")
-    assert print(p) == print(u)
+    assert u.name == p.name
+    assert u.key_attribute_names == p.key_attribute_names
+    assert u.description == p.description
+    assert u.external_id == p.external_id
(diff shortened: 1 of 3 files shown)


tests/unit/test_create_dataset.py
```diff
@@ -0,0 +1,30 @@
...
+
+    datasets_url = f"http://localhost:9100/api/versioned/v1/datasets"
+    dataset_url = f"http://localhost:9100/api/versioned/v1/datasets/1"
+
+    responses.add(responses.POST, datasets_url, json=dataset_config, status=204)
+    responses.add(responses.GET, dataset_url, json=dataset_config)
+
+    u = unify.create_dataset(dataset_config)
+    p = unify.datasets.by_resource_id("1")
+    assert print(p) == print(u)
```
pcattori: Printed representations are often for convenience and don't guarantee equality. If you want to make sure they are the same dataset, compare their important properties:

```suggestion
    assert u.id == p.id
    assert u.externalId == p.externalId
    assert u.name == p.name
    assert u.keyAttributeNames == p.keyAttributeNames
    assert u.description == p.description
```

---

Ideally, we could implement the `__eq__` function for some of our resources to make the comparison just `assert u == p`, but we don't yet have designs for what it *really* means to be the same dataset (e.g. what if its the same dataset at 2 different point in time after the `externalId` has been changed. should those be equal or not?)

tests/unit/test_create_dataset.py
```diff
@@ -0,0 +1,34 @@
...
+
+    datasets_url = f"http://localhost:9100/api/versioned/v1/datasets"
+    dataset_url = f"http://localhost:9100/api/versioned/v1/datasets/1"
+
+    responses.add(responses.POST, datasets_url, json=dataset_config, status=204)
+    responses.add(responses.GET, dataset_url, json=dataset_config)
+
+    u = unify.create_dataset(dataset_config)
+    p = unify.datasets.by_resource_id("1")
+    assert u.id == p.id
```
pcattori: This is causing tests to fail. I think should be:

```suggestion
    assert u.resource_id == p.resource_id
```

pcattori: @mollysacks you can [run tests locally](https://tamr-unify-python-client.readthedocs.io/en/latest/contributor-guide.html#run-tests) to make sure your PR will pass the automated testing.

PR #158, opened 2019-06-26
Fix #156 - Dataset profile can be out of date.

### LINKED ISSUE #156: Fetch profile, even if out-of-date
# 🙋 feature request

Currently, a dataset profile is automatically updated before returning, if the profile is out of date.
The client should just give me the profile in its current state, and let me decide whether to refresh it.

## 🤔 Expected Behavior

dataset.profile() returns a profile with `out_of_date=True`

## 😯 Current Behavior

dataset.profile() runs a job to update the profile, doesn't return until the job finishes.

nbateshaus:
# ↪️ Pull Request

Fix #156 - Fetch DatasetProfile, even if out of date.

This changes the behavior of Dataset.profile() - it no longer refreshes the profile.
This adds DatasetProfile.refresh() to replace the previous behavior.
This adds Dataset.create_profile() to create a profile if none exists.

## 💻 Examples

```python
        auth = UsernamePasswordAuth("username", "password")
        client = Client(auth)

        dataset = client.datasets.by_resource_id("3")
        op = dataset.create_profile()
        assert op.successful()
        profile = dataset.profile()
        if not profile.is_up_to_date:
                op = profile.refresh()
                assert op.successful()
```

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/models/dataset/resource.py
+++ b/tamr_unify_client/models/dataset/resource.py
@@ -104,27 +104,37 @@ class Dataset(BaseResource):
         op = Operation.from_json(self.client, op_json)
         return op.apply_options(**options)
 
-    def profile(self, **options):
-        """Returns up to date profile information for a dataset, re-profiling if not up to date.
+    def profile(self):
+        """Returns profile information for a dataset.
+
+        If profile information has not been generated, call create_profile() first.
+        If the returned profile information is out-of-date, you can call refresh() on the returned
+        object to bring it up-to-date.
 
         :param ``**options``: Options passed to underlying :class:`~tamr_unify_client.models.operation.Operation` .
         :return: Dataset Profile information.
         :rtype: :class:`~tamr_unify_client.models.dataset_status.DatasetProfile`
         """
-
         profile_json = self.client.get(self.api_path + "/profile").successful().json()
-        info = DatasetProfile.from_json(
+        return DatasetProfile.from_json(
             self.client, profile_json, api_path=self.api_path + "/profile"
         )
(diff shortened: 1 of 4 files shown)


tamr_unify_client/models/dataset/resource.py
```diff
@@ -104,27 +104,37 @@ def refresh(self, **options):
...
         return op.apply_options(**options)
 
-    def profile(self, **options):
-        """Returns up to date profile information for a dataset, re-profiling if not up to date.
+    def profile(self):
+        """Returns profile information for a dataset.
+
+        If profile information has not been generated, call create_profile() first.
+        If profile information is out-of-date, call refresh() on the returned
+        object to bring it up-to-date.
```
Nelson-Andrade: According to the ticket shouldn't we just be fetching the dataset in its current state as opposed to calling a refresh() on it?

tests/unit/test_dataset_profile.py
```diff
@@ -1,36 +1,119 @@
-import json
```
Nelson-Andrade: Seems like this line is necessary considering there is json within this file.

tamr_unify_client/models/dataset_profile.py
```diff
@@ -64,14 +65,24 @@ def attribute_profiles(self) -> list:
         """
         return self._data.get("attributeProfiles")
 
+    def refresh(self, **options):
+        """Brings this profile up-to-date if needed.
```
nbateshaus: Brings the profile up-to-date on the server; user will need to re-fetch the profile to get the up-to-date one.

PR #179, opened 2019-07-17
Get input datasets

juliamcclellan:
#178


tamr_unify_client/models/project/resource.py
```diff
@@ -110,6 +111,15 @@ def add_source_dataset(self, dataset):
         ).successful()
         return response
 
+    def source_datasets(self):
```
nbateshaus: we should use the same terminology as the API, so this should be called `input_datasets`

PR #197, opened 2019-07-19
Published clusters configuration

### LINKED ISSUE #184: Support for getting published cluster configurations

juliamcclellan:
Fix #184

### DIFF
--- a/tamr_unify_client/models/project/cluster_configuration.py
+++ b/tamr_unify_client/models/project/cluster_configuration.py
@@ -1,0 +1,44 @@
+from tamr_unify_client.models.base_resource import BaseResource
+
+
+class PublishedClustersConfiguration(BaseResource):
+    """
+    The configuration of published clusters in a project.
+
+    See https://docs.tamr.com/reference#the-published-clusters-configuration-object
+    """
+
+    @classmethod
+    def from_json(cls, client, data, api_path):
+        return super().from_data(client, data, api_path)
+
+    @property
+    def relative_id(self):
+        """:type: str"""
+        # api_path is alias when it exists, and relative_id when it does not.
+        # this distinction is useful for things like refreshing a unified dataset,
+        # where using the relative_id would hit
+        #   /datasets/{id}:refresh
+        # rather than
+        #   /projects/{id}/unifiedDataset:refresh.
+        # Since cluster configurations don't currently have that kind of aliasing,
+        # using api_path is always correct.
+        # If configurations ever get aliased, this will need to be updated.
+        # This is confusing; there's an RFC for suggestions to improve this
+        # #64 https://github.com/Datatamer/unify-client-python/issues/64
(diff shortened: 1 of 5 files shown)


tests/unit/test_published_clusters.py
```diff
@@ -17,15 +58,15 @@ def test_published_clusters():
...
+    _unified_dataset_json = {
         "id": "unify://unified-data/v1/datasets/8",
         "name": "Project_1_unified_dataset",
         "version": "10",
         "relativeId": "datasets/8",
         "externalId": "Project_1_unified_dataset",
     }
 
-    published_clusters_json = {
+    _published_clusters_json = {
```
charlottemoremen: why the underscore before "published"?

juliamcclellan: A lot of the existing tests do that for internal json variables

nbateshaus: The leading underscore is meaningful: 
https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables
"Use one leading underscore only for non-public methods and instance variables."

tamr_unify_client/models/project/mastering.py
```diff
@@ -85,6 +85,20 @@ def published_clusters(self):
...
+
+        :returns: The returned JSON body, as specified in the
+            `Public Docs for Cluster Configurations
+            <https://docs.tamr.com/reference#the-published-clusters-configuration-object>`_.
+        :rtype: dict
+        """
+        return (
+            self.client.get(self.api_path + "/publishedClustersConfiguration")
+            .successful()
+            .json()
```
nbateshaus: Even though it's a stupid class with only one member, we should provide a Python class for this.

juliamcclellan: Sad

tamr_unify_client/models/project/cluster_configuration.py
```diff
@@ -0,0 +1,25 @@
...
+    def from_json(cls, client, data, api_path):
+        return super().from_data(client, data, api_path)
+
+    @property
+    def relative_id(self):
+        """:type: str"""
+        return self.api_path
+
+    @property
+    def time_to_live(self):
```
nbateshaus: Should be `versions_time_to_live`

tamr_unify_client/models/project/cluster_configuration.py
```diff
@@ -0,0 +1,25 @@
...
+
+class PublishedClustersConfiguration(BaseResource):
+    @classmethod
+    def from_json(cls, client, data, api_path):
+        return super().from_data(client, data, api_path)
+
+    @property
+    def relative_id(self):
+        """:type: str"""
+        return self.api_path
```
nbateshaus: This needs an explanatory comment like the one in `attribute/resource.py`

PR #203, opened 2019-07-22
Attribute collection and type changes

### LINKED ISSUE #172: attribute's collection.py has a get_json method, even though it shouldn't

### LINKED ISSUE #175: Attribute's collection.py's init method uses self._data, and shouldn't, because we should be fetching fresh data when streaming
# 🐛 bug report

## 🌍 Your Environment

| Software          | Version(s) |
| ----------------- | ---------- |
| tamr-unify-client |
| Tamr Unify server |
| Python            |
| Operating System  |

juliamcclellan:
Fix #172 
Fix #175 

Doing a few things:
- getting rid of json attached to `AttributeCollection`s
- making `AttributeType` no longer inherit from `BaseResource` as it doesn't have an API endpoint
- creating a new `SubAttribute` class (not a `BaseResource`) that is a property of other `Attribute`s

### DIFF
--- a/tamr_unify_client/models/attribute/subattribute.py
+++ b/tamr_unify_client/models/attribute/subattribute.py
@@ -3,14 +3,11 @@ class SubAttribute:
     An attribute which is itself a property of another attribute.
     See https://docs.tamr.com/reference#attribute-types
 
-    :param client: Delegate underlying API calls to this client.
-    :type: :class:`~tamr_unify_client.Client`
     :param data: JSON data representing this attribute
     :type: :py:class:`dict`
     """
 
-    def __init__(self, client, data):
-        self.client = client
+    def __init__(self, data):
         self._data = data
 
     @property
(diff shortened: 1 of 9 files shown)


tamr_unify_client/models/attribute/subattribute.py
```diff
@@ -0,0 +1,45 @@
...
+    An attribute which is itself a property of another attribute.
+    See https://docs.tamr.com/reference#attribute-types
+
+    :param client: Delegate underlying API calls to this client.
+    :type: :class:`~tamr_unify_client.Client`
+    :param data: JSON data representing this attribute
+    :type: :py:class:`dict`
+    """
+
+    def __init__(self, client, data):
```
pcattori: I don't think `client` is necessary here. Usually we need it to fetch the resource itself or any subresources, but this isn't a resource and has no fetchable subresources. Could we remove `client` (including docstring) from this file?

PR #206, opened 2019-07-23
Dataset usage

### LINKED ISSUE #181: Support for seeing a dataset's usage
`datasets/{id}/usage`

juliamcclellan:
Fix #181

### DIFF
--- a/tamr_unify_client/models/dataset/usage.py
+++ b/tamr_unify_client/models/dataset/usage.py
@@ -1,0 +1,37 @@
+from tamr_unify_client.models.base_resource import BaseResource
+from tamr_unify_client.models.dataset.use import DatasetUse
+
+
+class DatasetUsage(BaseResource):
+    """
+    The usage of a dataset and its downstream dependencies.
+
+    See https://docs.tamr.com/reference#retrieve-downstream-dataset-usage
+    """
+
+    @classmethod
+    def from_json(self, client, resource_json, api_path):
+        return super().from_data(client, resource_json, api_path)
+
+    @property
+    def relative_id(self):
+        """:type: str"""
+        return self.api_path
+
+    @property
+    def usage(self):
+        """:type: :class:`~tamr_unify_client.models.dataset.use.DatasetUse`"""
+        return DatasetUse(self.client, self._data.get("usage"))
+
+    @property
+    def dependencies(self):
+        """:type: list[:class:`~tamr_unify_client.models.dataset.use.DatasetUse`]"""
+        deps = self._data.get("dependencies")
+        return [DatasetUse(self.client, dep) for dep in deps]
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__module__}."
+            f"{self.__class__.__qualname__}("
+            f"alias={self.api_path!r})"
+        )
--- a/tamr_unify_client/models/dataset/use.py
+++ b/tamr_unify_client/models/dataset/use.py
@@ -1,0 +1,57 @@
+from tamr_unify_client.models.project.step import ProjectStep
+
+
+class DatasetUse:
+    """
+    The use of a dataset in project steps. This is not a `BaseResource` because it has no API path
+        and cannot be directly retrieved or modified.
+
+    See https://docs.tamr.com/reference#retrieve-downstream-dataset-usage
+
+    :param client: Delegate underlying API calls to this client.
+    :type client: :class:`~tamr_unify_client.Client`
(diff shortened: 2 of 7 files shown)


tamr_unify_client/models/dataset/usage.py
```diff
@@ -0,0 +1,39 @@
+from tamr_unify_client.models.dataset.use import DatasetUse
+
+
+class DatasetUsage:
+    """
+    The usage of a dataset and its downstream dependencies.
+
+    See https://docs.tamr.com//reference#retrieve-downstream-dataset-usage
```
charlottemoremen: this is super nitpicky, but "See "https://docs.tamr.com//reference#retrieve-downstream-dataset-usage""

juliamcclellan: Doesn't seem to make any difference in the docs

tamr_unify_client/models/dataset/usage.py
```diff
@@ -0,0 +1,39 @@
...
+
+    :param client: Delegate underlying API calls to this client.
+    :type client: :class:`~tamr_unify_client.Client`
+    :param data: The JSON body containing usage information.
+    :type data: :py:class:`dict`
+    :param alias: API alias for this usage.
+    :type alias: str
+    """
+
+    def __init__(self, client, data, alias):
```
nbateshaus: Warrants a comment about why this is not a BaseResource

tamr_unify_client/models/dataset/use.py
```diff
@@ -0,0 +1,56 @@
+from tamr_unify_client.models.project.step import ProjectStep
+
+
+class DatasetUse:
+    """
+    The usage of a dataset in project steps.
```
nbateshaus: Confusing that this is DatasetUse, but the comment is about usage.

tamr_unify_client/models/dataset/use.py
```diff
@@ -0,0 +1,56 @@
...
+
+    See https://docs.tamr.com/reference#retrieve-downstream-dataset-usage
+
+    :param client: Delegate underlying API calls to this client.
+    :type client: :class:`~tamr_unify_client.Client`
+    :param data: The JSON body containing usage information.
+    :type data: :py:class:`dict`
+    """
+
+    def __init__(self, client, data):
```
nbateshaus: Warrants a comment about why this is not a BaseResource

tamr_unify_client/models/project/step.py
```diff
@@ -0,0 +1,68 @@
...
+
+    See https://docs.tamr.com/reference#retrieve-downstream-dataset-usage
+
+    :param client: Delegate underlying API calls to this client.
+    :type client: :class:`~tamr_unify_client.Client`
+    :param data: The JSON body containing project step information.
+    :type data: :py:class:`dict`
+    """
+
+    def __init__(self, client, data):
```
nbateshaus: Warrants a comment about why this is not a BaseResource

tamr_unify_client/models/project/step.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def project_name(self):
+        """:type: str"""
+        return self._data.get("projectName")
+
+    @property
+    def type(self):
+        """One of:
+            ``"SCHEMA_MAPPING"``
```
nbateshaus: This list will change out from under us, so we shouldn't embed it here.
Stated another way, Python Client doesn't control the contract for what the valid values are for this string. Since Python Client doesn't control it, Python Client can't actually assert what the valid values are. We should refer to the contract, rather than replicate it.

juliamcclellan: This should be changed in the project file then too. I can do that during the refactor.

tamr_unify_client/models/dataset/usage.py
juliamcclellan: It should have been I guess.

tamr_unify_client/models/dataset/usage.py
```diff
@@ -0,0 +1,36 @@
...
+
+    See https://docs.tamr.com/reference#retrieve-downstream-dataset-usage
+    """
+
+    @classmethod
+    def from_json(self, client, resource_json, api_path):
+        return super().from_data(client, resource_json, api_path)
+
+    @property
+    def relative_id(self):
```
nbateshaus: Doesn't `DatasetUsage` get this for free from `BaseResource`?

PR #208, opened 2019-07-23
Refresh published cluster stats

### LINKED ISSUE #202: /projects/{project}/publishedClusterStats:refresh
# 🙋 feature request

`/projects/{project}/publishedClusterStats:refresh`

juliamcclellan:
Fix #202 

Yet another way to update published clusters, not sure why it's necessary


tests/unit/test_published_clusters.py
```diff
@@ -69,6 +69,17 @@ def test_refresh_ids(self):
         self.assertEqual(op.resource_id, self._operations_json["id"])
         self.assertTrue(op.succeeded())
 
+    @responses.activate
+    def test_refresh_stats(self):
```
charlottemoremen: could this be consolidated into one line?

charlottemoremen: lines 74 and 75

juliamcclellan: Yes, it would just be an extremely long line. I think I'll actually move the base URL to be a class variable since it's used in other tests as well.

tests/unit/test_published_clusters.py
```diff
@@ -60,15 +60,44 @@ def test_published_clusters_configuration(self):
...
+        )
+        responses.add(responses.GET, datasets_url, json=self._datasets_json)
         responses.add(responses.POST, refresh_url, json=self._operations_json)
 
         p = Project(self.unify, self._project_config_json).as_mastering()
-        op = p.refresh_published_cluster_ids()
+        d = p.published_cluster_ids()
+
+        op = d.refresh()
         self.assertEqual(op.resource_id, self._operations_json["id"])
```
nbateshaus: <3 that the validation didn't change here

PR #214, opened 2019-07-26
Stream dataset records as updates

### LINKED ISSUE #205: implement a fetch records from dataset function into python client
# 🙋 feature request

 implement a fetch records from dataset function into python client to easily transfer records from dataset across instances

## 🤔 Expected Behavior

ds1.from_records(ds2.records()) to replicate records from ds2 to ds1 

## 😯 Current Behavior

transferring a dataset from one instance of Unify to another, can be achieved by following scripts:
ds1 = unify1.datasets.by_external_id(ds1_id)
ds2 = unify2.datasets.by_external_id(ds2_id)
def records_to_updates(pk, records):
        for record in records:
            yield {
                "action": "CREATE",
                "recordId": record[pk],
                "record": record,
            }
ds2.update_records(records_to_updates(ds1.key_attribute_names[0], ds1.records())) 

## 🔦 Context

 I am building "Cluster Time Travel Demo" using airflow and need to transfer data from one instance to another, the cluster time travel  demo will show the cluster change over time.  I need to rebuild the demo in our new demo instance. this can not be achieved by simple replicate project because time stamps will not be kept. to replicate this demo, I need to redo the mastering workflow for multiple time points. we decided to use airflow to help automate the whole process.

juliamcclellan:
Fix #205 

Given datasets `d1` and `d2`, the following code works:
`d2.upsert_records(d1.records(), d1_primary_key_name)`

### DIFF
--- a/tamr_unify_client/models/dataset/resource.py
+++ b/tamr_unify_client/models/dataset/resource.py
@@ -80,6 +80,24 @@ class Dataset(BaseResource):
             .json()
         )
 
+    def upsert_records(self, records, primary_key_name, **json_args):
+        """Converts the records into update commands and calls :func:`~tamr_unify_client.models.dataset.resource.Dataset.update_records`
+
+        :param records: The records to update, as dictionaries.
+        :type records: iterable[dict]
+        :param primary_key_name: The name of the primary key for these records, which must be a key in each record dictionary.
+        :type primary_key_name: str
+        :param `**json_args`: Arguments to pass to the JSON `dumps` function, as documented `here <https://simplejson.readthedocs.io/en/latest/#simplejson.dumps>`_.
+            Some of these, such as `indent`, may not work with Unify.
+        :return: JSON response body from the server.
+        :rtype: dict
+        """
+        updates = (
+            {"action": "CREATE", "recordId": record[primary_key_name], "record": record}
+            for record in records
+        )
+        return self.update_records(updates, **json_args)
+
     def refresh(self, **options):
         """Brings dataset up-to-date if needed, taking whatever actions are required.
         :param ``**options``: Options passed to underlying :class:`~tamr_unify_client.models.operation.Operation` .
--- a/tamr_unify_client/models/dataset/resource.py
+++ b/tamr_unify_client/models/dataset/resource.py
@@ -55,8 +55,10 @@ class Dataset(BaseResource):
         alias = self.api_path + "/attributes"
         return AttributeCollection(self.client, alias)
 
-    def update_records(self, records, **json_args):
+    def _update_records(self, updates, **json_args):
         """Send a batch of record creations/updates/deletions to this dataset.
+        You probably want to use :func:`~tamr_unify_client.models.dataset.resource.Dataset.upsert_records`
+        or :func:`~tamr_unify_client.models.dataset.resource.Dataset.delete_records` instead.
 
         :param records: Each record should be formatted as specified in the `Public Docs for Dataset updates <https://docs.tamr.com/reference#modify-a-datasets-records>`_.
         :type records: iterable[dict]
@@ -65,23 +67,22 @@ class Dataset(BaseResource):
         :returns: JSON response body from server.
         :rtype: :py:class:`dict`
         """
-
-        def _stringify_updates(updates):
-            for update in updates:
-                yield json.dumps(update, **json_args).encode("utf-8")
+        stringified_updates = (
+            json.dumps(update, **json_args).encode("utf-8") for update in updates
+        )
 
         return (
             self.client.post(
                 self.api_path + ":updateRecords",
                 headers={"Content-Encoding": "utf-8"},
-                data=_stringify_updates(records),
+                data=stringified_updates,
             )
             .successful()
             .json()
         )
 
     def upsert_records(self, records, primary_key_name, **json_args):
-        """Converts the records into update commands and calls :func:`~tamr_unify_client.models.dataset.resource.Dataset.update_records`
+        """Creates or updates the specified records.
 
         :param records: The records to update, as dictionaries.
         :type records: iterable[dict]
@@ -96,10 +97,10 @@ class Dataset(BaseResource):
             {"action": "CREATE", "recordId": record[primary_key_name], "record": record}
             for record in records
         )
-        return self.update_records(updates, **json_args)
+        return self._update_records(updates, **json_args)
 
     def delete_records(self, records, primary_key_name):
-        """Converts the records into delete commands and calls :func:`~tamr_unify_client.models.dataset.resource.Dataset.update_records`
+        """Deletes the specified records.
 
         :param records: The records to delete, as dictionaries.
         :type records: iterable[dict]
@@ -112,7 +113,7 @@ class Dataset(BaseResource):
             {"action": "DELETE", "recordId": record[primary_key_name]}
             for record in records
         )
-        return self.update_records(updates)
+        return self._update_records(updates)
 
     def refresh(self, **options):
         """Brings dataset up-to-date if needed, taking whatever actions are required.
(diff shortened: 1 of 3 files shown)


tamr_unify_client/models/dataset/resource.py
```diff
@@ -66,20 +68,60 @@ def update_records(self, records, **json_args):
...
                 self.api_path + ":updateRecords",
                 headers={"Content-Encoding": "utf-8"},
-                data=_stringify_updates(records),
+                data=_stringify_updates(updates),
             )
             .successful()
             .json()
         )
 
+    def upsert_records(self, records, key_name, **json_args):
```
pcattori: nit: I think `primary_key` might be a clearer name than `key_name` since "primary key" is a Unify datasets concept, where as "key" and "name" are generic. Could be `primary_key_name` if we want to be more explicit.

tamr_unify_client/models/dataset/resource.py
```diff
@@ -66,20 +68,60 @@ def update_records(self, records, **json_args):
...
+        :type records: iterable[dict]
+        :param key_name: The name of the primary key for these records, which must be a key in each record dictionary.
+        :type key_name: str
+        :param `**json_args`: Arguments to pass to the JSON `dumps` function, as documented `here <https://simplejson.readthedocs.io/en/latest/#simplejson.dumps>`_.
+            Some of these, such as `indent`, may not work with Unify.
+        :return: JSON response body from the server.
+        :rtype: dict
+        """
+
+        def _to_updates(updates):
```
pcattori: Can simplify via:

```python
updates = (
  {
    "action": "CREATE",
    "recordId": record[primary_key],
    "record": record,
  } for record in records
)

return self._update_records(updates, **json_args)
```

tamr_unify_client/models/dataset/resource.py
```diff
@@ -66,20 +68,60 @@ def update_records(self, records, **json_args):
...
+
+        :param records: The records to delete, as dictionaries.
+        :type records: iterable[dict]
+        :param key_name: The name of the primary key for these records, which must be a key in each record dictionary.
+        :type key_name: str
+        :return: JSON response body from the server.
+        :rtype: dict
+        """
+
+        def _to_deletes(deletes):
```
pcattori: similar simplification can be made here.

tests/unit/test_dataset_records.py
```diff
@@ -72,17 +72,61 @@ def create_callback(request, snoop, status):
...
+
+    @staticmethod
+    def records_to_deletes(records):
+        return [{"action": "DELETE", "recordId": i + 1} for i in range(len(records))]
+
     @staticmethod
     def records_to_updates(records):
         return [
-            {"action": "CREATE", "recordId": str(i), "record": records[i]}
+            {"action": "CREATE", "recordId": i + 1, "record": records[i]}
```
pcattori: when iterating over index + element in a Python iterable, use `enumerate`:

```python
{"action": "CREATE", "recordId": i + 1, "record": record}
for i, record in enumerate(records)
```

tests/unit/test_dataset_records.py
```diff
@@ -72,17 +72,61 @@ def create_callback(request, snoop, status):
...
+        response = dataset.delete_records(self._records_json, "attribute1")
+        self.assertEqual(response, self._response_json)
+        self.assertEqual(snoop["payload"], TestDatasetRecords.stringify(deletes, False))
+
+    @staticmethod
+    def records_to_deletes(records):
+        return [{"action": "DELETE", "recordId": i + 1} for i in range(len(records))]
+
     @staticmethod
     def records_to_updates(records):
```
pcattori: this should be renamed to `records_to_creates`

tests/unit/test_dataset_records.py
```diff
@@ -72,17 +72,61 @@ def create_callback(request, snoop, status):
...
+            responses.POST, records_url, partial(create_callback, snoop=snoop)
+        )
+
+        response = dataset.delete_records(self._records_json, "attribute1")
+        self.assertEqual(response, self._response_json)
+        self.assertEqual(snoop["payload"], TestDatasetRecords.stringify(deletes, False))
+
+    @staticmethod
+    def records_to_deletes(records):
+        return [{"action": "DELETE", "recordId": i + 1} for i in range(len(records))]
```
pcattori: same comment about using `enumerate`

tamr_unify_client/models/dataset/resource.py
pcattori: Even if you don't go with this^, the argument should be renamed from `updates` to `records`:

```python
def _to_updates(records):
```

tests/unit/test_dataset_records.py
```diff
@@ -72,18 +72,64 @@ def create_callback(request, snoop, status):
...
+        )
+
+        response = dataset.delete_records(self._records_json, "attribute1")
+        self.assertEqual(response, self._response_json)
+        self.assertEqual(snoop["payload"], TestDatasetRecords.stringify(deletes, False))
+
+    @staticmethod
+    def records_to_deletes(records):
+        return [
+            {"action": "DELETE", "recordId": i + 1} for i, record in enumerate(records)
```
pcattori: if you want enumeration to start at `1` instead of `0`, you can do:

```suggestion
            {"action": "DELETE", "recordId": i + 1} for i, record in enumerate(records, start=1)
```

tests/unit/test_dataset_records.py
```diff
@@ -72,18 +72,64 @@ def create_callback(request, snoop, status):
...
+        ]
+
     @staticmethod
     def records_to_updates(records):
         return [
-            {"action": "CREATE", "recordId": str(i), "record": records[i]}
-            for i in range(len(records))
+            {"action": "CREATE", "recordId": i + 1, "record": records[i]}
```
pcattori: same thing about `enumerate(..., start=1)`

tamr_unify_client/models/dataset/resource.py
```diff
@@ -65,21 +67,54 @@ def update_records(self, records, **json_args):
...
+        :return: JSON response body from the server.
+        :rtype: dict
+        """
+        updates = (
+            {"action": "CREATE", "recordId": record[primary_key_name], "record": record}
+            for record in records
+        )
+        return self._update_records(updates, **json_args)
+
+    def delete_records(self, records, primary_key_name):
```
pcattori: There's another use-case where I want to delete a record, but I only know its ID. In this case, would be nice to have a `dataset.delete_records_by_id(record_ids)`. @juliamcclellan that can be part of this PR or we can make an issue for it and do it later.

tests/unit/test_dataset_records.py
juliamcclellan: There are really so many python features I know nothing about

tamr_unify_client/models/dataset/resource.py
```diff
@@ -65,21 +67,65 @@ def update_records(self, records, **json_args):
...
+        """Deletes the specified records.
+
+        :param records: The records to delete, as dictionaries.
+        :type records: iterable[dict]
+        :param primary_key_name: The name of the primary key for these records, which must be a key in each record dictionary.
+        :type primary_key_name: str
+        :return: JSON response body from the server.
+        :rtype: dict
+        """
+        updates = (
```
pcattori: this method implementation can be simplified via:

```python
delete_ids = (record['primary_key_name'] for record in records)
return self.delete_records_by_id(delete_ids)
```

tamr_unify_client/models/dataset/resource.py
```diff
@@ -65,21 +67,62 @@ def update_records(self, records, **json_args):
         :returns: JSON response body from server.
         :rtype: :py:class:`dict`
         """
-
-        def _stringify_updates(updates):
-            for update in updates:
-                yield json.dumps(update, **json_args).encode("utf-8")
+        stringified_updates = (
```
nbateshaus: This breaks streaming :-(

pcattori: Even though this looks like a tuple-comprehension (which isn't a thing), it's a [generator expression](https://www.python.org/dev/peps/pep-0289/#the-details). So records are still streamed and never materialized in RAM.

PR #230, opened 2019-08-01
Delete all records from a dataset

### LINKED ISSUE #220: Delete a dataset's records
Dataset has an endpoint `DELETE /datasets/{id}/records` which deletes all records in a dataset. This deletion would not be covered by a delete for `BaseResource` or `BaseCollection`, and is more general than the existing dataset record deletion functions.

juliamcclellan:
# ↪️ Pull Request

Fix #220 

First delete endpoint, so not sure if it should wait for v0.9.0.

I don't know how to unit test it more. It does work on a real unify instance.

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/dataset/resource.py
+++ b/tamr_unify_client/dataset/resource.py
@@ -123,6 +123,16 @@ class Dataset(BaseResource):
         updates = ({"action": "DELETE", "recordId": rid} for rid in record_ids)
         return self._update_records(updates)
 
+    def delete_all_records(self):
+        """Removes all records from the dataset.
+
+        :return: HTTP response from the server
+        :rtype: :class:`requests.Response`
+        """
+        path = self.api_path + "/records"
+        response = self.client.delete(path).successful()
+        return response
+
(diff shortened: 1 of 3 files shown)


tamr_unify_client/dataset/resource.py
```diff
@@ -123,6 +123,16 @@ def delete_records_by_id(self, record_ids):
         updates = ({"action": "DELETE", "recordId": rid} for rid in record_ids)
         return self._update_records(updates)
 
+    def delete_all_records(self):
```
pcattori: The API models this as a `DELETE` on the `/records` subcollection of a dataset, so this should be `dataset.records().delete_all()` or something like that. We may need to remodel `/records` as a `RecordCollection` rather than just giving back a stream of records in order to accomplish this.

pcattori: On second thought: `/records` isn't a proper collection yet (e.g. you can't get a member of the collection via APIs), so let's keep this code as is.

PR #236, opened 2019-08-06
Getting attributes from a collection

### LINKED ISSUE #235: Attribute collection by_name streams all attributes
The `AttributeCollection` function `by_name` works by streaming the collection and searching for the name. However, it could just call GET on `f"{self.api_path}/{attribute_name}"`

juliamcclellan:
# ↪️ Pull Request

Fix #235 

Currently, to retrieve a specific attribute, an `AttributeCollection` streams the attributes and locates the one with the right name. This change is to directly use the API endpoints that get specific records instead (for attributes, name is equivalent to resource ID).

## ✔️ PR Todo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## 0.9.0-dev
+  **BUG FIXES**
+   - [#235](https://github.com/Datatamer/unify-client-python/issues/235) Making `AttributeCollection` retrieve attributes directly instead of by streaming
 
 ## 0.8.0
   **BREAKING CHANGES**
--- a/tamr_unify_client/attribute/collection.py
+++ b/tamr_unify_client/attribute/collection.py
@@ -23,7 +23,7 @@ class AttributeCollection(BaseCollection):
         :returns: The specified attribute.
         :rtype: :class:`~tamr_unify_client.attribute.resource.Attribute`
         """
-        return self.by_name(resource_id)
+        return super().by_resource_id(self.api_path, resource_id)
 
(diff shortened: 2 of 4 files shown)


tamr_unify_client/attribute/collection.py
```diff
@@ -23,7 +23,7 @@ def by_resource_id(self, resource_id):
         :returns: The specified attribute.
         :rtype: :class:`~tamr_unify_client.attribute.resource.Attribute`
         """
-        return self.by_name(resource_id)
```
pcattori: This is a backwards incompatible change. One idea @nbateshaus and I discussed earlier on is to try the newer API, but if it fails with a `404`, to fallback on the old, less nice way of doing things. This means a user on an older version can still use the latest version of python client to interact with their instance, while letting users with new versions take advantage of the newer APIs.

Let's discuss this during standup.

pcattori: This API has existed since the inception of `versioned/v1` APIs, so it is not a backwards incompatible change.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 
   **BUG FIXES**
   - [#212](https://github.com/Datatamer/unify-client-python/issues/212) Sped up slow running tests
+  - [#235](https://github.com/Datatamer/unify-client-python/issues/235) Making `AttributeCollection` retrieve attributes directly instead of by streaming
```
pcattori: @juliamcclellan can you rebase on `master` so that this goes into `0.9.0-dev`? we already bumped the version on `master`

PR #237, opened 2019-08-06
Cluster configuration builder

### LINKED ISSUE #226: Update published cluster configurations with put
Create a way to modify an existing cluster configuration object and `PUT` the result to commit the updates

charlottemoremen:
# ↪️ Pull Request

Fix #226 

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/mastering/published_cluster/configuration.py
+++ b/tamr_unify_client/mastering/published_cluster/configuration.py
@@ -1,3 +1,5 @@
+import copy
+
 from tamr_unify_client.base_resource import BaseResource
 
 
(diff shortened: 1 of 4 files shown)


tamr_unify_client/mastering/published_cluster/configuration_builder.py
```diff
@@ -0,0 +1,48 @@
+import copy
+
+from tamr_unify_client.mastering.published_cluster.configuration import (
+    PublishedClustersConfiguration,
+)
+
+
+class PublishedClustersConfigurationBuilder:
```
pcattori: we should place this builder in the same file as its built type (`PublishedClustersConfiguration`).

tests/unit/test_published_clusters.py
```diff
@@ -1,3 +1,5 @@
+from functools import partial
+import json
 from unittest import TestCase
```
pcattori: Note: In the future use [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html) instead of `unittest.TestCase`

pcattori:
Code looks good. One design question I have: what's the benefit we are getting from having a separate builder class? Why not just have `with_versions_time_to_live` as a method on `PublishedClustersConfiguration` that returns a new `PublishedClustersConfiguration` (similar to immutable interfaces)?

@nbateshaus ^ curious what your thoughts are.

tamr_unify_client/mastering/published_cluster/configuration.py
```diff
@@ -42,3 +44,46 @@ def __repr__(self):
...
+
+class PublishedClustersConfigurationBuilder:
+    """A builder object to modify an existing published cluster configuration.
+
+    :param cluster_configuration: The cluster configuration to modify.
+    :type cluster_configuration: :class:`~tamr_unify_client.mastering.published_cluster.configuration.PublishedClustersConfiguration`
+    """
+
+    def __init__(self, cluster_configuration):
+        self._data = copy.deepcopy(cluster_configuration._data)
```
nbateshaus: This design is fine for mutable builder, but we should refactor for immutable builders.

PR #169, opened 2019-07-09
Fix #168 - Add support for Project Attributes

nbateshaus:
# ↪️ Pull Request

Fix #168 - Add support for Project Attributes

## 💻 Examples

```
id_attribute = project.attributes.create({"name":"id", "type":{"baseType":"STRING"},"isNullable":False})
for attribute in project.attributes:
    do_something_with(attribute)
```

## ✔️ PR Todo

- [x] Added/updated testing for this change
- [x] Included links to related issues/PRs
- [x] Update relevant [docs](https://github.com/Datatamer/unify-client-python/tree/master/docs) + docstrings
- [x] Update the [CHANGELOG](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md) under the current `-dev` version:
  - Add changelog entries under any that apply: **BREAKING CHANGES**, **NEW FEATURES**, **BUG FIXES**.
  - Changelog entry format: `[#<issue number>](<link to issue>) <change description>`

PR #247, opened 2019-08-08
Create a dataset from a dataframe

### LINKED ISSUE #67: Create new dataset
# 💬 RFC

Trying to upload new data. I would prefer to create a new dataset, but the API doesn't support that. Instead, I need to use `Dataset.update_records`.

I would like to be able to create a new dataset.

## 🔦 Context

Concerns with updating into an existing dataset:
- Does not provide isolation of batches of data. With different datasets, I can map a given record to a particular load operation.
- Different datasets allow for partitioning, a common data management technique. Partitioning would allow for deletion of a single dataset and allow the dataset name to be an indicator of which data has already been uploaded (imagine that I have a system of uploading data daily, with a name based on the date).
- Records within a single dataset require unique `recordId` values, which requires additional management.
- Making a larger dataset causes later operations to be more costly. This further hurts the performance in the context of #66, for example.

It appears that the server's API does support the creation of new datasets, but this isn't provided in the python client. https://docs.tamr.com/reference#create-dataset

## 💁 Possible Solution

```
def create_dataset(unify, dataset_config):
    """
    Create a dataset in Unify
    
    :param unify: Unify Client
    :param dataset_config: Dataset Configuration
    :return: the created Dataset
    """
    from tamr_unify_client.models.dataset.resource import Dataset
    data = unify.post(unify.datasets.api_path, json=dataset_config).successful().json()
    return Dataset(unify, data, data["relativeId"])
```

juliamcclellan:
# ↪️ Pull Request
#169
Fix #67 

## ✔️ PR Todo


docs/conf.py
```diff
@@ -52,6 +52,7 @@
 intersphinx_mapping = {
     "https://docs.python.org/": None,
     "requests": ("http://docs.python-requests.org/en/master/", None),
+    "pandas": ("https://pandas.pydata.org/pandas-docs/stable", None),
```
pcattori: can you provide a screenshot showing that links to `pandas` docs from our docs work?

tests/unit/test_create_dataset.py
```diff
@@ -1,33 +1,234 @@
...
 
 from tamr_unify_client import Client
 from tamr_unify_client.auth import UsernamePasswordAuth
+from tamr_unify_client.dataset.collection import CreationError
 
-auth = UsernamePasswordAuth("username", "password")
-unify = Client(auth)
 
+class DatasetCreationTest(TestCase):
+    def setUp(self):
```
pcattori: Could you use a [`pytest` fixture](https://docs.pytest.org/en/latest/fixture.html) here instead of `unittest.TestCase`? We already use `pytest` as our testing framework, so I'd like to use that throughout the codebase (I know there are places where we use `unittest` too, but we should strive to unify things under 1 testing framework).

tamr_unify_client/dataset/collection.py
```diff
@@ -89,4 +91,119 @@ def create(self, creation_spec):
...
+        try:
+            dataset.delete()
+        except HTTPError as e:
+            raise CreationError(CreationError.DELETION_ERROR, e)
+        raise CreationError(error_type, error)
+
     # super.__repr__ is sufficient
+
+
+class CreationError(RuntimeError):
```
pcattori: Python official docs recommend that [all user-defined exceptions be derived from `Exception`](https://docs.python.org/3/library/exceptions.html#Exception):

```suggestion
class CreationError(Exception):
```

tamr_unify_client/dataset/collection.py
```diff
@@ -89,4 +91,119 @@ def create(self, creation_spec):
...
+    """An error from :func:`~tamr_unify_client.dataset.collection.DatasetCollection.create_from_dataframe`"""
+
+    NOT_CREATED = 0
```
pcattori: Rather than having an enumeration of different error types as instance properties, we should make more custom exception classes that inherit from `CreationError`

docs/conf.py
juliamcclellan: Screen Shot 2019-08-12 at 9 49 12 AM

tests/unit/test_create_dataset.py
juliamcclellan: I think you and Danger should talk about this because you seem to disagree

PR #248, opened 2019-08-08
Delete an Attribute Mapping 

### LINKED ISSUE #221: Delete an attribute mapping
As attribute mappings will not be `BaseResource`s, their delete endpoint will need to be handled separately.

There should be a `delete` function in `AttributeMapping`, and a `delete_by_resource_id` in `AttributeMappingCollection`

charlottemoremen:
# ↪️ Fixes #221 

## ✔️ PR Todo

- [ ] Update relevant [docs](https://github.com/Datatamer/unify-client-python/tree/master/docs) + docstrings
- [ ] Update the [CHANGELOG](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md) under the current `-dev` version:


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## 0.9.0-dev
   **NEW FEATURES**
+  -[#221] (https://github.com/Datatamer/unify-client-python/issues/221) delete an attribute mapping
```
juliamcclellan: I think there needs to be a space between the - and [ and not between ] and (
(check how the markdown renders on github)

tamr_unify_client/project/attribute_mapping/collection.py
```diff
@@ -55,3 +55,14 @@ def create(self, creation_spec):
         """
         data = self.client.post(self.api_path, json=creation_spec).successful().json()
         return AttributeMapping(data)
+
+    def delete_by_resource_id(self, resource_id):
+        """delete an attribute mapping using its Resource ID.
+        :param resource_id: the resource ID of the mapping to be deleted.
+        :type resource_id: str
+        :returns: the status code (200, 400, etc.)
+        :rtype: HTTP status code
```
juliamcclellan: What you're returning here isn't just the status code, but the response itself, so type :class:\`requests.Response\`

tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -64,6 +64,11 @@ def resource_id(self):
         spliced = self.relative_id.split("attributeMappings/")[1]
         return spliced
 
+    def delete(self, url, client):
```
juliamcclellan: This shouldn't need to take in url and client. It seems like right now attribute mapping does not have client and api path properties, which is an issue here. The attribute mapping `__init__` should take these in to fix this (api path can be optional, as it is the same as the mapping's relative ID--do this similarly to how base resource does)

Also needs a docstring

tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -64,6 +64,11 @@ def resource_id(self):
...
+        delete_url = url + "/" + self.resource_id
```
juliamcclellan: This is the same as `self.relative_id`

tests/unit/test_attribute_mapping.py
```diff
@@ -37,6 +45,17 @@ def test_resource(self):
         expected = self.mappings_json["unifiedAttributeName"]
         self.assertEqual(expected, test.unified_attribute_name)
 
+    @responses.activate
+    def test_delete(self):
+        general_url = (
+            "http://localhost:9100/api/versioned/v1/projects/4/attributeMappings"
+        )
+        specific_url = "http://localhost:9100/api/versioned/v1/projects/4/attributeMappings/19629-12"
```
juliamcclellan: You could base this off of `general_url` (`specific_url = general_url + "/19629-12"`)

tests/unit/test_attribute_mapping.py
```diff
@@ -37,6 +45,17 @@ def test_resource(self):
...
+        responses.add(responses.DELETE, specific_url, json=self.delete_json)
```
juliamcclellan: `DELETE` endpoints have empty response bodies, so you shouldn't pass a `json` parameter. Instead, use `status=204`

tests/unit/test_attribute_mapping.py
```diff
@@ -37,6 +45,17 @@ def test_resource(self):
...
+        delete_map = AttributeMapping(self.delete_json)
+        final_response = delete_map.delete(general_url, self.unify)
+        print(final_response)
```
juliamcclellan: You aren't asserting anything in this test--you should check that the status is correct.

tests/unit/test_attribute_mapping.py
```diff
@@ -49,3 +68,16 @@ def test_resource(self):
         "unifiedDatasetName": "Charlotte_unified_dataset",
         "unifiedAttributeName": "surname",
     }
+
+    delete_json = {
```
juliamcclellan: This seems identical to the already defined mappings_json

tests/unit/test_attribute_mapping_collection.py
```diff
@@ -38,6 +38,16 @@ def test_by_relative_id(self):
...
 
+    @responses.activate
+    def test_delete(self):
+        url = "http://localhost:9100/api/versioned/v1/projects/4/attributeMappings"
+        responses.add(responses.GET, url, json=self.mappings_json)
+        deleteColl = AttributeMappingCollection(self.unify, url)
+        url2 = "http://localhost:9100/api/versioned/v1/projects/4/attributeMappings/19629-12"
+        responses.add(responses.DELETE, url2, json=self.delete_json)
+        response = deleteColl.delete_by_resource_id("19629-12")
+        print(response)
```
juliamcclellan: Same general comments for this test as the other

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
...
   - [#218](https://github.com/Datatamer/unify-client-python/issues/218) Delete a `BaseResource`
-
```
juliamcclellan: The newline is also required for markdown rendering

tamr_unify_client/project/attribute_mapping/resource.py
charlottemoremen: self.relative_Id = "projects/4/attributeMappings/19629-12"
delete_url ="http://localhost:9100/api/versioned/v1/projects/4/attributeMappings/19629-12"

juliamcclellan: `delete_url` should just need to be `projects/4/attributeMappings/19629-12` because the client will have a base path of `http://localhost:9100/api/versioned/v1`

tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -5,8 +5,9 @@ class AttributeMapping:
     (ex: /projects/1/attributeMappings/1), but these types of URLs do not exist for attribute mappings
     """
 
-    def __init__(self, data):
+    def __init__(self, data, client=None):
```
juliamcclellan: For consistency with the other resources, the order here should be `client, data, alias=None`. This would go into the "breaking changes" section of the changelog

tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -64,6 +65,14 @@ def resource_id(self):
         spliced = self.relative_id.split("attributeMappings/")[1]
         return spliced
 
+    @property
+    def delete(self):
+        """:return: :class `requests.response`
```
juliamcclellan: Needs a more detailed comment

tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -64,6 +65,14 @@ def resource_id(self):
...
+        """
+        delete_url = self.client.origin + self.client.base_path + self.relative_id
+        response = self.client.delete(delete_url).successful()
```
juliamcclellan: So when you make a request with the client, it will handle all the base path stuff for you. All you need to do is pass it the relative_id, and it will concatenate that with the base path information.

tests/unit/test_attribute_mapping.py
```diff
@@ -37,6 +45,14 @@ def test_resource(self):
         expected = self.mappings_json["unifiedAttributeName"]
         self.assertEqual(expected, test.unified_attribute_name)
 
+    @responses.activate
+    def test_delete(self):
+        specific_url = "http://localhost:9100/api/versioned/v1/projects/4/attributeMappings/19629-12"
+        responses.add(responses.DELETE, specific_url, status=204)
+        delete_map = AttributeMapping(self.mappings_json, self.unify)
+        final_response = delete_map.delete
+        self.assertEqual(str(final_response), "<Response [204]>")
```
juliamcclellan: It's safer to do this without converting to a string (since in theory it could be something entirely wrong that happens to have the same repr)--so assert `final_response.status_code` is `204`

tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -64,6 +65,14 @@ def resource_id(self):
         spliced = self.relative_id.split("attributeMappings/")[1]
         return spliced
 
+    @property
```
juliamcclellan: This shouldn't be a property, it's a function

tests/unit/test_attribute_mapping_collection.py
```diff
@@ -38,6 +38,18 @@ def test_by_relative_id(self):
...
+    def test_delete(self):
+        general_url = (
+            "http://localhost:9100/api/versioned/v1/projects/4/attributeMappings"
+        )
+        responses.add(responses.GET, general_url, json=self.mappings_json)
+        deleteColl = AttributeMappingCollection(self.unify, general_url)
+        specific_url = general_url + "/19629-12"
+        responses.add(responses.DELETE, specific_url, status=204)
+        response = deleteColl.delete_by_resource_id("19629-12")
+        self.assertEqual(str(response), "<Response [204]>")
```
juliamcclellan: Same comment here

Issue #252, opened 2019-08-11
Builder pattern / implementation


pcattori:
# 💬 RFC

We have PRs that implement the builder pattern for some resources (#237 , #249 , #250 ). We should decide on overall design for how we want to apply the builder pattern in our codebase.

E.g.
- Should the builder be accessible via `MyClass.builder()` (which returns a `MyClassBuilder` instance)?
- Should the build use REST verbs (e.g. `.put()`) OR should it use domain-specific verbs (e.g. `create`, `upsert`, etc...)
- Should the builder just be able to build instance via `.build()` and then any communication with the server is done via client? (e.g. `mc = MyClass.builder().with_option(param).build(); client.put(mc)`

nbateshaus:
1. Yes, builder should be accessible via `MyClass.builder()` on all classes for which Builder is appropriate.
3. `Builder.build()` should be the one place where the logic for going from MyClass's API to the Tamr server's JSON API is encapsulated. That is, `MyClassBuilder.build()` should produce the JSON that can be sent to the server.
2. We should use REST verbs; they're already ubiquitous in the code, and, as this is a client for a REST API, they are the best description for what's actually going on. `MyClassBuilder.put()` should be:
```
def put(self):
    data = self.client.put(self.api_path, json=self.build()).successful().json()
    return MyClass(self.client, data)
```

juliamcclellan:
@charlottemoremen, @Nelson-Andrade, and I have two possible options:

The first is to have a `BaseBuilder` class which handles overriding properties and has an abstract `put`. Each resource would have a concrete builder with `with_property` functions that created new instances of the builder class.

```
class X:
    def __init__(self, client, data, alias):
        self.client = client
        self._data = data
        self.api_path = alias

    def a(self):
        return self._data.get("a")

    def build():
        return XBuilder(self)

class BaseBuilder:
    def __init__(self, contained, **kwargs):
        self._contained = contained
        self.__dict__.update(**kwargs)

    def __getattr__(self, name):
        return getattr(self._contained, name)

    @abstractmethod
    def _build(self):
        pass

    @abstractmethod
    def put(self, resource_class):
        new_data = self.client.put(self.api_path, json=self._build()).successful().json()
        return resource_class(self.client, new_data, self.api_path)

class XBuilder(Builder):
    def with_a(self, new_a):
        return XBuilder(self, a=new_a)

    def _build(self):
        return {
            "a": self.a
        }

    def put(self):
        return super().put(X)
```

[...]

PR #262, opened 2019-08-13
Attribute configuration and dataset specs

### LINKED ISSUE #222: Update a dataset with put
Create a way to modify an existing dataset and `PUT` the result to commit the updates

### LINKED ISSUE #225: Update an attribute configuration with put
Create a way to modify an existing attribute configuration and `PUT` the result to commit the updates

juliamcclellan:
# ↪️ Pull Request

An option for #252 

Fix #222 
Fix #225 

## 💻 Examples

```

unify = Client(UsernamePasswordAuth("username", "password"), "ip")
p = unify.projects.by_resource_id("1")
configs = p.attribute_configurations()

keep_ml = ["email", "last", "zip"]

for config in configs:
    if config.attribute_name not in keep_ml:
        config.spec().with_enabled_for_ml(False).put()
```

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/dataset/resource.py
+++ b/tamr_unify_client/dataset/resource.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+
 import simplejson as json
 
 from tamr_unify_client.attribute.collection import AttributeCollection
@@ -262,6 +264,14 @@ class Dataset(BaseResource):
 
         return [DatasetURI(self.client, uri) for uri in resources]
 
+    def spec(self):
+        """Returns this dataset's spec.
+
+        :return: The spec of this dataset.
+        :rtype: :class:`~tamr_unify_client.dataset.resource.DatasetSpec`
+        """
+        return DatasetSpec.of(self)
+
     @property
     def __geo_interface__(self):
         """Retrieve a representation of this dataset that conforms to the Python Geo Interface.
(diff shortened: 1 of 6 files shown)


tamr_unify_client/dataset/resource.py
```diff
@@ -434,3 +444,88 @@ def _geo_attr_names():
             "polygon",
             "multiPolygon",
         }
+
+
+class DatasetSpec:
+    """A representation of the server view of a dataset."""
+
+    def __init__(self, client, data, api_path):
```
pcattori: nit: I suggest we use `client, api_path, data` ordering to group together the things that won't change (i.e. `client` and `api_path`) together, and keep `data` at the end.

even more minor nit: Additionally, you could consider ordering the assigments within the `__init__` method in the same order as they are declared in the `__init__` function signature.

tamr_unify_client/project/attribute_configuration/resource.py
```diff
@@ -72,3 +82,123 @@ def __repr__(self):
...
+    """A representation of the server view of an attribute configuration."""
+
+    def __init__(self, client, data, api_path):
+        self._data = data
+        self.client = client
+        self.api_path = api_path
+
+    @staticmethod
+    def of(attribute_configuration):
+        """Creates an attribute configuration spec from an attribute configuration.
```
pcattori: The code itself is repeated, but quite short, which I think is fine. However, these docstrings are pretty long and repeated. Having to keep these in sync could become annoying. Maybe we should document how `*Spec` classes look (e.g. centralize the docstrings for `*Spec.of` into 1 place) and just reference that doc from the `*Spec` class docstring.

Then we would only need to document the `with_*` with type information.

pcattori: We should discuss this and come to a consensus that we document in the RFC #252

tamr_unify_client/dataset/resource.py
nbateshaus: We should use the same order as `Dataset.__init__()`, otherwise users will get it wrong every single time.

pcattori: @nbateshaus users only need to use `.spec()` so they never have to order it the same.

tamr_unify_client/dataset/resource.py
```diff
@@ -434,3 +444,88 @@ def _geo_attr_names():
...
+
+        :param data: The data for the new spec.
+        :type data: dict
+        :return: The new spec.
+        :rtype: :class:`~tamr_unify_client.dataset.resource.DatasetSpec`
+        """
+        return DatasetSpec(self.client, data, self.api_path)
+
+    def to_dict(self):
+        """Returns a copy of this spec's internal representation.
```
nbateshaus: The contract here isn't really 'this spec's internal representation' - rather, the contract should be that `to_dict()` will return a `dict` that conforms to the API of `PUT datasets/{id}`

tamr_unify_client/dataset/resource.py
```diff
@@ -434,3 +444,88 @@ def _geo_attr_names():
...
+        :return: A new spec.
+        :rtype: :class:`~tamr_unify_client.dataset.resource.DatasetSpec`
+        """
+        return self.from_data({**self._data, "description": new_description})
+
+    def with_tags(self, new_tags):
+        """Creates a new spec with the same properties, updating tags.
+
+        :param new_tags: The new tags.
+        :type new_tags: list[str]
```
nbateshaus: `list[str]` isn't our contract; it's the contract of the API. My preference would be to explicitly defer to the API contract, but the reality is that contract is supposed to not change, so it's both low risk and user friendly to replicate it here.

tamr_unify_client/dataset/resource.py
```diff
@@ -434,3 +444,88 @@ def _geo_attr_names():
...
+class DatasetSpec:
+    """A representation of the server view of a dataset."""
+
+    def __init__(self, client, data, api_path):
+        self._data = data
+        self.client = client
+        self.api_path = api_path
+
+    @staticmethod
+    def of(dataset):
```
nbateshaus: nit: it is more readable (and more copy-pasteable) if the argument to `of()` is something generic, like `resource`.
```
def of(resource):
    return DatasetSpec(resource.client, deepcopy(resource._data), resource.api_path)
```

tests/unit/test_attribute_configuration.py
```diff
@@ -57,22 +59,56 @@ def test_resource_from_json(self):
...
+        responses.add(responses.GET, config_url, json=self._ac_json)
+        responses.add_callback(
+            responses.PUT, config_url, partial(create_callback, snoop=snoop_dict)
+        )
+        configs = AttributeConfigurationCollection(self.unify, self._alias)
+        config = configs.by_resource_id(self._attribute_id)
+
+        temp_spec = config.spec().with_attribute_role("SUM_ATTRIBUTE")
+        new_config = (
+            temp_spec.with_enabled_for_ml(False)
```
nbateshaus: nit: fluent style wants even the first `.with_*` on a separate line:
```
    temp_spec
        .with_enabled_for_ml(False)
        .with_similarity_function("ABSOLUTE_DIFF")
        .with_tokenizer("BIGRAM")
        .put()
```

tests/unit/test_attribute_configuration.py
```diff
@@ -57,22 +59,56 @@ def test_resource_from_json(self):
...
+            .with_tokenizer("BIGRAM")
+            .put()
+        )
+
+        self.assertEqual(new_config._data, self._updated_ac_json)
+        self.assertEqual(json.loads(snoop_dict["payload"]), self._updated_ac_json)
+        self.assertEqual(config._data, self._ac_json)
+
+        # checking that intermediate didn't change
+        self.assertTrue(temp_spec.to_dict()["enabledForMl"])
```
nbateshaus: 👍

tests/unit/test_dataset.py
```diff
@@ -26,6 +27,40 @@ def test_delete(self):
...
+            responses.PUT, self._url, partial(create_callback, snoop=snoop_dict)
+        )
+
+        dataset = self.unify.datasets.by_resource_id("1")
+
+        temp_spec = dataset.spec().with_description(
+            self._updated_dataset_json["description"]
+        )
+        new_dataset = (
+            temp_spec.with_external_id(self._updated_dataset_json["externalId"])
```
nbateshaus: nit: fluent style

tamr_unify_client/dataset/resource.py
juliamcclellan: What @nbateshaus and I decided:
If a property of a resource is a complex type (like an attribute type), it should have its own spec class, and a user should provide an instance of this spec to `with_*`.
If a property is not a complex type (e.g. a string), it would not have a spec class, and the user would provide the simple representation directly.
If the property is a list/set of resources, the user would provide a list of these resources, represented as determined by the above.

A tag is represented as a string, so there should not be a resource or spec class for a tag. Consequently, a user should just provide a list of strings to `with_tags`.

tests/unit/test_attribute_configuration.py
juliamcclellan: Black, however, does not.

PR #280, opened 2019-08-16
Fix #221 - Delete an attribute mapping

### LINKED ISSUE #221: Delete an attribute mapping

charlottemoremen:
# ↪️ Pull Request

Fix #221 - Delete an attribute mapping

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/project/attribute_mapping/resource.py
+++ b/tamr_unify_client/project/attribute_mapping/resource.py
@@ -8,8 +8,10 @@ class AttributeMapping:
     (ex: /projects/1/attributeMappings/1), but these types of URLs do not exist for attribute mappings
     """
 
-    def __init__(self, data):
+    def __init__(self, client, data, alias=None):
         self._data = data
+        self.client = client
+        self.api_path = alias or self.relative_id
 
     @property
     def id(self):
@@ -74,6 +76,14 @@ class AttributeMapping:
         :rtype: :class:`~tamr_unify_client.project.attribute_mapping.resource.AttributeMappingSpec`
         """
         return AttributeMappingSpec.of(self)
+
+    def delete(self):
+        """delete this attribute mapping.
+        :return: HTTP response from the server
+        :rtype: :class `requests.Response`
+        """
+        response = self.client.delete(self.api_path).successful()
+        return response
 
     def __repr__(self):
         return (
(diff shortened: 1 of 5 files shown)


tamr_unify_client/project/attribute_mapping/resource.py
```diff
@@ -8,8 +8,10 @@ class AttributeMapping:
     (ex: /projects/1/attributeMappings/1), but these types of URLs do not exist for attribute mappings
     """
 
-    def __init__(self, data):
+    def __init__(self, client, data, alias=None):
```
pcattori: do `AttributeMappings` have meaningful aliases? or can they only be accessed those resources only be reference via 1 canonical URL? Trying to figure out if `alias` makes sense to include here or not.

tests/unit/test_attribute_mapping.py
```diff
@@ -1,51 +1,71 @@
...
 
 from tamr_unify_client.project.attribute_mapping.resource import AttributeMapping
 
 
-class TestAttributeMapping(TestCase):
-    def test_resource(self):
-        test = AttributeMapping(self.mappings_json)
+@pytest.fixture
+def client():
+    from tamr_unify_client import Client
```
pcattori: Are there cyclic import issues with putting these `import`s at the top of the file? dynamic imports are usually only idiomatic to avoid cyclic imports

tamr_unify_client/project/attribute_mapping/resource.py
nbateshaus: At present, attribute mappings cannot be aliased.
I'll remove the alias.

tests/unit/test_attribute_mapping.py
nbateshaus: No issue with cyclic imports; I was motivated purely by removing namespace clutter at the file level.

tests/unit/test_attribute_mapping_collection.py
```diff
@@ -1,172 +1,181 @@
...
-        )
-        map_collection.create(spec.to_dict())
-
-        self.assertEqual(snoop_dict["payload"], self.create_json)
-
-    create_json = {
-        "relativeInputAttributeId": "datasets/6/attributes/suburb",
+@pytest.fixture
+def client():
+    from tamr_unify_client import Client
```
pcattori: could you move these imports to be at the top so that we are consistent with the other test file?

nbateshaus: Thanks, I missed that one.

PR #290, opened 2019-08-29
Handle empty HTTP 204 responses for already-completed operations.

### LINKED ISSUE #293: Better handling for HTTP 204 on already up-to-date operations
# 🙋 feature request

When doing something that starts an operation (e.g. a dataset `.refresh()`), if everything that would be produced by that operation is already up-to-date, the server will return an HTTP 204 with no content. The various bits of client code that do these things try to read the empty content as JSON to create an `Operation`, but raise a `json.JSONDecodeError` instead (because an empty document is not valid for the JSON data model).

## 🤔 Expected Behavior

```
published_clusters = project.published_clusters()
op = published_clusters.refresh()
assert op.successful
```

## 😯 Current Behavior

```
published_clusters = project.published_clusters()
try:
    published_clusters.refresh()
except json.JSONDecodeError:
    # In the event that the dataset we want to refresh is already up-to-date,
    # the API will return an HTTP 200 with an empty body, causing a JSONDecodeError.
    # We can safely ignore this error.
    pass
```

## 💁 Possible Solution

Whenever the server gives back an HTTP 204, create a dummy or No-Op operation to represent that.

## 🔦 Context

Putting the exception handling in is counter-intuitive and makes scripts verbose and hard to read.

It's hard to know exactly which things might return HTTP 204, so often the first I learn that I missed a spot is when I'm trying to run something and it fails unexpectedly.

Only after running into this a few times does one learn that `JSONDecodeError` means the operation was successful and nothing needed to be done on the server.

## 💻 Examples

See above.

nbateshaus:
# ↪️ Pull Request

Fix #293 Better handling for HTTP 204 on already up-to-date operations

Handle empty HTTP 204 responses for already-completed operations.

This creates a NOOP Operation to represent these cases.

## 💻 Examples

Before:
```
published_clusters = project.published_clusters()
try:
    published_clusters.refresh()
except json.JSONDecodeError:
    # In the event that the dataset we want to refresh is already up-to-date,
    # the API will return an HTTP 200 with an empty body, causing a JSONDecodeError.
    # We can safely ignore this error.
    pass
```
After:
```
published_clusters = project.published_clusters()
op = published_clusters.refresh()
assert op.successful
```

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/operation.py
+++ b/tamr_unify_client/operation.py
@@ -18,6 +18,48 @@ class Operation(BaseResource):
     def from_json(cls, client, resource_json, api_path=None):
         return super().from_data(client, resource_json, api_path)
 
+    @classmethod
+    def from_response(cls, client, response):
+        """
+        Handle idiosyncrasies in constructing Operations from Tamr responses.
+        When a Tamr API call would start an operation, but all results that would be
+        produced by that operation are already up-to-date, Tamr returns `HTTP 204 No Content`
+
+        To make it easy for client code to handle these API responses without checking
+        the response code, this method will either construct an Operation, or a
+        dummy `NoOp` operation representing the 204 Success response.
+
+        :param client: Delegate underlying API calls to this client.
+        :type client: :class:`~tamr_unify_client.Client`
+        :param response: HTTP Response from the request that started the operation.
+        :type response: :class:`requests.Response`
+        :return: Operation
+        :rtype: :class:`~tamr_unify_client.operation.Operation`
+        """
+        if response.status_code == 204:
+            # Operation was successful, but the response contains no content.
+            # Create a dummy operation to represent this.
+            _never = "0000-00-00T00:00:00.000Z"
+            _description = """Tamr returned HTTP 204 for this operation, indicating that all
+                results that would be produced by the operation are already up-to-date."""
+            resource_json = {
+                "id": "-1",
+                "type": "NOOP",
+                "description": _description,
+                "status": {
+                    "state": "SUCCEEDED",
+                    "startTime": _never,
+                    "endTime": _never,
+                    "message": "",
+                },
+                "created": {"username": "", "time": _never, "version": "-1"},
+                "lastModified": {"username": "", "time": _never, "version": "-1"},
+                "relativeId": "operations/-1",
+            }
+        else:
+            resource_json = response.json()
+        return Operation.from_json(client, resource_json)
+
     def apply_options(self, asynchronous=False, **options):
         """Applies operation options to this operation.
 
--- a/tests/unit/test_operation.py
+++ b/tests/unit/test_operation.py
@@ -1,0 +1,95 @@
+from urllib.parse import urljoin
+
+import pytest
+from requests import HTTPError
+import responses
+
+
+from tamr_unify_client.operation import Operation
+
+
+@pytest.fixture
+def client():
+    from tamr_unify_client import Client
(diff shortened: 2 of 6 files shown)


tamr_unify_client/operation.py
```diff
@@ -18,6 +18,52 @@ class Operation(BaseResource):
...
+    def from_response(cls, client, response):
+        """
+        Handle idiosyncrasies in constructing Operations from Tamr responses.
+        When a Tamr API call would start an operation, but all results that would be
+        produced by that operation are already up-to-date, Tamr returns:
+            HTTP 204 No Content
+
+        To make it easy for client code to handle these API responses without checking
+        the response code, this method will either construct an Operation, or a
+        dummy NoOp operation representing the fact the 204 Success response.
```
juliamcclellan: Remove "the fact"

tamr_unify_client/operation.py
```diff
@@ -18,6 +18,52 @@ class Operation(BaseResource):
...
+
+        :param :class:`~tamr_unify_client.Client` client: Delegate underlying API calls to this client.
```
juliamcclellan: This doesn't render properly in the docs, it needs a line with `:param client: ...` and one with `:type client: :class: ...`

tamr_unify_client/operation.py
```diff
@@ -18,6 +18,52 @@ class Operation(BaseResource):
...
+        if response.status_code == 204:
+            # Operation was successful, but the response contains no content.
+            # Create a dummy operation to represent this.
+            resource_json = {
+                "id": "-1",
+                "type": "NOOP",
+                "description": "NoOp",
+                "status": {
+                    "state": "SUCCEEDED",
+                    "startTime": "0000-00-00T00:00:00.000Z",
```
juliamcclellan: It might be better to define the time string above as it's repeated several times

tamr_unify_client/operation.py
```diff
@@ -18,6 +18,46 @@ class Operation(BaseResource):
...
+        :rtype: :class:`~tamr_unify_client.operation.Operation`
+        """
+        if response.status_code == 204:
+            # Operation was successful, but the response contains no content.
+            # Create a dummy operation to represent this.
+            _never = "0000-00-00T00:00:00.000Z"
+            resource_json = {
+                "id": "-1",
+                "type": "NOOP",
+                "description": "NoOp",
```
pcattori: @nbateshaus we should add a prose-y description along the lines of the docstring for this function, rather than `NoOP`. That will enable users to debug more effectively without having to look at the docs.

tests/unit/test_operation.py
```diff
@@ -0,0 +1,96 @@
...
+
+    op2a = op2.apply_options(asynchronous=True)
+    assert op2a.succeeded
+
+    op2w = op2a.wait()
+    assert op2w.succeeded
+
+    with pytest.raises(HTTPError):
+        op2p = op2w.poll()
+        assert not op2p.succeeded
```
pcattori: Is this `assert` actually reached? Looks to me like this line could be removed without changing the test.

tamr_unify_client/operation.py
nbateshaus: Done.

tests/unit/test_operation.py
nbateshaus: It's there to 'use' `op2p`, but we can remove `op2p` since `pytest.raises` will fail if the exception isn't raised.

Issue #300, opened 2019-10-03
Google-style docstrings + Python 3 type annotations


pcattori:
The [`sphinx.ext.napoleon`](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) extension allows for more readable docstrings and the use of type annotations to enrich docstrings with type info. See [this section of the napoleon docs](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#type-annotations) for details.

We were previously using `.rst`-formatted docstrings, but we are now moving away from `.rst` as outlined in #299 .

## Example

`sphinx` will produce the same doc HTML for the following 2 examples:

Old `.rst`-formatted style:

```python
def func(arg1, arg2):
    """Summary line. Extended description of function.

        :param arg1: Description of arg1
        :type arg1: :class:`~tamr_unify_client.mastering.project.MasteringProject`
        :param arg2 str: Description of arg2
        :returns: Description of return value
        :rtype: bool
    """
    return True
```

New style via `sphinx.ext.napoleon`:

```python
def func(arg1: MasteringProject, arg2: str) -> bool:
    """Summary line.

    Extended description of function.

    Args:
        arg1: Description of arg1
        arg2: Description of arg2

    Returns:
        Description of return value

    """
    return True
```

## Migration plan

Actions:
- Whenever a docstring or corresponding function is updated, try to convert the docstring to Google-style docstring and include type annotations.
- All new docstrings should be written in Google-style. All new functions with docstrings should include type annotations

## How to migrate

- [Examples of Google-style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
- [`typing`](https://docs.python.org/3/library/typing.html) module from Python standard library describes type annotations in detail

If you get stuck, ask for help by leaving a comment on this issue.

pcattori:
Obsoleted by #385

PR #303, opened 2019-10-10
POC for dataclasses

pcattori:
# ↪️ Pull Request

Relates to #300 (specifically, adds [`sphinx-autodoc-typehints`](https://github.com/agronholm/sphinx-autodoc-typehints))

Proof of Concept for using `@dataclass` for resources in our codebase.
See [this video](https://www.youtube.com/watch?v=nwjWOaxWMes) for motivation / explanation of `dataclasses` module.
[In this video](https://www.youtube.com/watch?v=T-TwcmT6Rcw), Ramond Hettinger (a Python core contributor) advocates for using `dataclasses`.

## 💻 Examples

See `SubAttribute.py`

Docs render as:
Screenshot_20191010_111749

## TODO
 - [X] Add the [official `dataclasses` backport](https://github.com/ericvsmith/dataclasses) for Python 3.6.

## ✔️ PR Todo

- [X] N/A ~Added/updated testing for this change~


tamr_unify_client/attribute/subattribute.py
```diff
@@ -1,43 +1,38 @@
+from copy import deepcopy
+from dataclasses import dataclass, field
+from typing import Any, Dict, Optional
+
+from tamr_unify_client.attribute.type import AttributeType
+
+SubAttributeJson = Dict[str, Any]
+
+@dataclass(frozen=True)
```
pcattori: `dataclass` implements `__init__`, `__repr__`, `__eq__` based on type annotations. `frozen=True` makes instances immutable.

tamr_unify_client/attribute/subattribute.py
```diff
@@ -1,43 +1,38 @@
...
+        is_nullable: If this sub-attribute can be null
     """
+    name: str
+    type: AttributeType
+    is_nullable: bool
+    _json: SubAttributeJson = field(repr=False)
+    description: Optional[str] = None
+
+    @staticmethod
+    def from_json(d: SubAttributeJson) -> 'SubAttribute':
```
pcattori: In Python 3.6, cyclic type references (e.g. the return type of this function refers to the type of the class implementing it) require using strings. See [PEP 563](https://www.python.org/dev/peps/pep-0563/) for how this can be made easier in Python 3.7

tamr_unify_client/attribute/subattribute.py
```diff
@@ -1,43 +1,38 @@
...
+    Args:
+        name: Name of sub-attribute
+        description: Description of sub-attribute
+        type: See https://docs.tamr.com/reference#attribute-types
+        is_nullable: If this sub-attribute can be null
     """
+    name: str
+    type: AttributeType
+    is_nullable: bool
+    _json: SubAttributeJson = field(repr=False)
```
abw333: can you explain this line and the following? i understand the previous 3

tamr_unify_client/attribute/subattribute.py
abw333: what's the policy for upgrading/supporting python versions?

tamr_unify_client/attribute/subattribute.py
```diff
@@ -1,43 +1,38 @@
...
+    description: Optional[str] = None
+
+    @staticmethod
+    def from_json(d: SubAttributeJson) -> 'SubAttribute':
+        _json = deepcopy(d)
+
+        dc = deepcopy(d)
+        dc['is_nullable'] = dc.pop('isNullable')
```
abw333: what's going on here? is it just renaming the key from jsonCase to python_case?

tamr_unify_client/attribute/subattribute.py
```diff
@@ -1,43 +1,38 @@
...
     """
+    name: str
+    type: AttributeType
+    is_nullable: bool
+    _json: SubAttributeJson = field(repr=False)
+    description: Optional[str] = None
+
+    @staticmethod
+    def from_json(d: SubAttributeJson) -> 'SubAttribute':
+        _json = deepcopy(d)
```
abw333: why do you need to maintain this property?

tamr_unify_client/attribute/subattribute.py
pcattori: this tells `dataclass` not to include the `_json` property when generating the `__repr__` implementation

tamr_unify_client/attribute/subattribute.py
abw333: cool. would perhaps be nice to abstract this into a superclass at some point

tamr_unify_client/attribute/subattribute.py
pcattori: we support 3.6+ for Python Client. Ubuntu 19 and many modern linux distros ship with 3.6, so this decision keeps life simple for DataOps folks @ Tamr

tamr_unify_client/attribute/subattribute.py
abw333: that's the leading underscore? what about the `foo = bar`? and `field(repr=False)`?

pcattori: no. I'm talking about `field(repr=False)`. `foo = bar` is syntax for default values in type annotations. leading underscore is just convention so users don't blindly rely on `_json`

tamr_unify_client/attribute/type.py
```diff
@@ -32,8 +30,10 @@ def inner_type(self):
     @property
     def attributes(self):
         """:type: list[:class:`~tamr_unify_client.attribute.subattribute.SubAttribute`]"""
+        from tamr_unify_client.attribute.subattribute import SubAttribute
```
Lordluen: This is for my own education, but why move the import from the file level to the function level? @pcattori

abw333: i think this is mostly to not run all that import logic unless it's needed

pcattori: `Attribute` and `SubAttribute` both depend on each other. If they each declare the other as imports at the file level, this will cause a [cyclic dependency](http://effbot.org/zone/import-confusion.htm#circular-imports). 

Previously, we got around this by having `SubAttribute` import `Attribute` at the function level, but conceptually `SubAttribute` depends much more heavily on `Attribute` than the other way around. So I changed this to reflect that. Specifically, now `SubAttribute` imports `Attribute` at the file level since it fundamentally depends on `Attribute`, whereas `Attribute` only depends on `SubAttribute` in 1 context (this one).

pcattori: to be clear: the code won't run successfully with cyclic imports. it will crash. so this is a correctness thing, not a perf thing

PR #304, opened 2019-10-10
Flask proxy

pcattori:
# ↪️ Pull Request

Utility for using `tamr-client` to proxy requests in a Flask app.

## 💻 Examples

```python
tamr = Client(auth, host)

passthru = proxy(tamr)
app = Flask(__name__)

@app.route("/api/endpoint/to/intercept")
def some_route(*args, **kwargs):
    # specially process this endpoint before passing it through to Tamr server
    print('this endpoint was called!')
    return passthru(*args, **kwargs)

# all other endpoints, passthru without doing anything special
app.route("/api/<path:path>", methods=["POST", "GET", "PUT", "DELETE"])(passthru)
```

### DIFF
--- a/examples/flask-proxy/app.py
+++ b/examples/flask-proxy/app.py
@@ -1,0 +1,23 @@
+from tamr_unify_client import Client
+from tamr_unify_client.auth import UsernamePasswordAuth
+
+from flask_proxy import flask_proxy
+
+auth = UsernamePasswordAuth("username", "password")
+tamr = Client(auth, host="my_host")
+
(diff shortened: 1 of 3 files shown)


pcattori:
@nbateshaus implemented as an example

pcattori:
Note: as is, this flask proxy does support streaming downloads, but not streaming uploads

pcattori:
This should go in [Datatamer/tamr-client-extensions](https://github.com/Datatamer/tamr-client-extensions) instead.

PR #312, opened 2019-12-13
Show response body for unsuccessful responses

### LINKED ISSUE #307: Client should print helpful messages when it raises exceptions
# 🙋 feature request

When a colleague tried to use the Tamr client to create a project, her terminal printed a 400 error for "Bad request," and of course Python pointed to the line number in her code. As a new hire, she didn't know what could be bad about the request, especially because the Tamr client structured most of the request for her. We eventually determined that her project spec JSON had the project type as "MASTERING" when Tamr expects "DEDUP." Surely the Tamr client can be improved to parse the error messages from Tamr and pass them on to the user, perhaps as a note to stdout or as the improved text of an exception.

## 🤔 Expected Behavior

In the above example, at least the Tamr client should make available some message to at least suggest that the project spec JSON might have invalid or missing values. That's a reasonable suggestion given that the only possibility for user error in this method is the project spec. At the most, the Tamr client can check the format and values of specs and other method arguments and return helpful errors.

## 😯 Current Behavior

Now, the user needs to either include exception handling for every use Tamr client usage, or read the documentation closely to intuit the problem. Both of these options are contrary to the goal (as I understand it) of the Tamr client: to make Tamr easier to use for our customers.

## 💁 Possible Solution

See expected behavior above.

pcattori:
# ↪️ Pull Request

Before, unsuccessful response would show:
1. HTTP Status code
2. Reason (High-level summary text)
3. URL

But much of the useful debugging data resides in the response body and
was not exposed to the user.

Now, we include the response body in the HTTPError message.

Fixes #307

## 💻 Examples

Everything after (and including) `Response body` is new. Body redacted for brevity.

```sh
>>> tamr.get('/api/dataset/datasets/1').successful()
Traceback (most recent call last):
  File "/Users/pedro/workspace/Tamr/tamr-client/tamr_unify_client/client.py", line 20, in successful
    response.raise_for_status()
  File "/Users/pedro/Library/Caches/pypoetry/virtualenvs/tamr-unify-client-CPhoWqYR-py3.6/lib/python3.6/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://10.10.0.201:9100/api/dataset/datasets/1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/pedro/workspace/Tamr/tamr-client/tamr_unify_client/client.py", line 24, in successful
    raise requests.exceptions.HTTPError(msg, response=response)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://10.10.0.201:9100/api/dataset/datasets/1
Response body:
{"status":404,"class":"com.tamr.common.except.ResourceNotFoundException","message":"Dataset with id 1 not found.","stackTrace":["com.tamr.dataset.service.DatasetService::getDataset::DatasetService.java::231","com.tamr.dataset.api.resources.BaseDatasetResource::lambda$getParentResource$0::BaseDatasetResource.java::71","com.tamr.persist.backend.postgres.PostgresBackend::callInSession::PostgresBackend.java::131","com.tamr.dataset.api.resources.BaseDatasetResource::getParentResource::BaseDatasetResource.java::69","com.tamr.dataset.api.resources.BaseDatasetResou...
...
...
...
...rg.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread::run::ReservedThreadExecutor.java::366","org.eclipse.jetty.util.thread.QueuedThreadPool::runJob::QueuedThreadPool.java::762","org.eclipse.jetty.util.thread.QueuedThreadPool$2::run::QueuedThreadPool.java::680","java.lang.Thread::run::Thread.java::745"],"causedBy":null,"service":"dataset"}
```

## ✔️ PR Todo

- [N/A] Added/updated testing for this change


pyproject.toml
```diff
@@ -25,17 +25,17 @@ python = "^3.6"
 requests = "^2.22"
 simplejson = "^3.16"
 dataclasses = "^0.6.0"
+pandas = "^0.25.3"
```
nbateshaus: This is a pretty heavyweight change.
In general, pandas is not *required* to use python-client; it's optional, only required if you use dataset.collection.create_from_dataframe()

tamr_unify_client/client.py
```diff
@@ -1,26 +1,31 @@
...
+        requests.exceptions.HTTPError: If an HTTP error is encountered (i.e. status code is 4xx or 5xx).
     """
-    self.raise_for_status()
-    return self
+    try:
+        response.raise_for_status()
+        return response
+    except requests.exceptions.HTTPError as e:
+        msg = str(e) + f"\nResponse body:\n{response.text}"
+        raise requests.exceptions.HTTPError(msg, response=response)
```
nbateshaus: It is kind of infuriating how non-standardized the Python `Exception` interface is.
My understanding was that the first argument is a message, which is intended to be a short summary text, and that further information is to be contained in other parts of the exception. This seems not only to be undocumented, but not the case - many Python Exceptions don't have a message (e.g. `OSError`).
`HTTPError` is only slightly less poorly documented than Exceptions in general; of note here is that, if no request is provided, `HTTPError.request` is filled in from `HTTPError.response.request`
So.
1. This information is already available in `e.response.text`; this is just promoting it up to the message. The message may become quite long as a result (e.g. HTTP 404 that returns an HTML body)
2. If the HTTPError contains a stack trace or other information, it will be lost in this construction.

pyproject.toml
pcattori: It was (and remains) a `dev` dependency, so this only affects people developing Python Client that need to run tests for `create_from_dataframe()`. This does not require `pandas` for users that install without dev deps (e.g. from PyPI)

tamr_unify_client/client.py
nbateshaus: To avoid the message `During handling of the above exception, another exception occurred:`, the `except:` block should complete without raising an exception, and the exception should be raised from outside the `try:` block.

nbateshaus: Or, `e` should be modified in-place and re-raised.

pcattori: @nbateshaus The original error is not lost since previous exceptions in the chain can be accessed via `e.__context__` or `e.__cause__`, but I agree that we might want to handle (1) better. But that puts the onus on the user to write a `try`/`catch` block for EVERY `successful()` call. Perhaps we can log the error in some way instead? Maybe just write a "tamr-client.err.log" file from `cwd` and stick it in there?

nbateshaus: Promoting `e.response.text` to `e.args[0]` is not a bad thing. I *wish* there were a standard place to put more information, but there doesn't seem to be; since this seems like crucial information, it seems right to put it there.

nbateshaus: I think all of this works fine if `e.args[0]` is modified in-place to append `e.response.text`, and e is re-thrown.

pcattori: The issue with storing the response body in the exception is that if the user did not already have a `try`/`catch` block, the response body will no be recoverable and they will have to rerun after modifying their code. Plus the ergonomics of surrounding EVERY tamr client call with a `try`/`catch` block are not great.

The issue with printing out the response body is that the response body is that it goes against the "short summary" purpose of the message and could be an obnoxious amount of info to print on the terminal.

A solution that avoids both of these issues is to modify this code so that the original, unaltered `HTTPError` is re-raised AFTER we log the response body to a `tamr-client.log` file.

@nbateshaus : thoughts on this approach^^^?

pcattori: caught up with @nbateshaus offline. will proceed by replacing code in this PR to log (not print) the response body. users can configure logging to be to be to stdout, a file, etc..

pcattori:
Superseded by #315

PR #315, opened 2019-12-19
Logging improvements

### LINKED ISSUE #307: Client should print helpful messages when it raises exceptions

pcattori:
# ↪️ Pull Request

Fixes #307
Supersedes #312 

- Logging response body for requests with HTTP error codes (4xx, 5xx)
- Adhere to [logging best practices for libraries](https://docs.python-guide.org/writing/logging/#logging-in-a-library)
- Rewrote logging docs and promoted to their own section (no longer in advanced usage)

Also:
- Progress on #300 for `client.py`

## 💻 Examples

```python
# script.py
import logging

logging.basicConfig(filename="script.log", level=logging.INFO)

# configure logging before other imports

auth = UsernamePasswordAuth("my username", "my password")
tamr = Client(auth, host="myhost")

for p in tamr.projects:
    print(p)

for d in tamr.datasets:
    print(d)

# should cause an HTTP error
tamr.get("/invalid/api/path").successful()
```
^^^ will log all API requests and the response body for the requests with HTTP error codes to `script.log`

## ✔️ PR Todo

- [N/A] Added/updated testing for this change


tamr_unify_client/client.py
```diff
@@ -1,62 +1,76 @@
...
 
-def successful(self):
-    """Checks that this response did not encounter an HTTP error (i.e. status code indicates success: 2xx, 3xx).
+def successful(response: requests.Response) -> requests.Response:
+    """Checks that this response did not encounter an HTTP error.
 
-    :raises :class:`requests.exceptions.HTTPError`: If an HTTP error is encountered.
-    :return: The calling response (i.e. ``self``).
-    :rtype: :class:`requests.Response`
+    HTTP error codes match 4xx or 5xx.
```
nbateshaus: This isn't quite true; we're delegating to `requests.Response.raise_for_status`, and it's fair to say so.

tamr_unify_client/client.py
```diff
@@ -1,62 +1,76 @@
...
 from tamr_unify_client.project.collection import ProjectCollection
 
-# monkey-patch Response.successful
+logger = logging.getLogger(__name__)
 
 
-def successful(self):
-    """Checks that this response did not encounter an HTTP error (i.e. status code indicates success: 2xx, 3xx).
+def successful(response: requests.Response) -> requests.Response:
+    """Checks that this response did not encounter an HTTP error.
```
nbateshaus: Per recent discussion, it's more clear to say that this will return the current response if no error occurred, or raise the stored error if an error did occur.

tamr_unify_client/client.py
```diff
@@ -1,62 +1,76 @@
...
-    :raises :class:`requests.exceptions.HTTPError`: If an HTTP error is encountered.
-    :return: The calling response (i.e. ``self``).
-    :rtype: :class:`requests.Response`
+    HTTP error codes match 4xx or 5xx.
+
+    Returns:
+        The response being checked.
+
+    Raises:
+        requests.exceptions.HTTPError: If an HTTP error is encountered.
```
nbateshaus: per `requests.Response.raise_for_status`, this can raise HTTPError or URLError
https://2.python-requests.org//en/v0.6.5/api/#requests.Response.raise_for_status

tamr_unify_client/client.py
```diff
@@ -1,62 +1,76 @@
...
     """
-    self.raise_for_status()
-    return self
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError as e:
+        logger.error(f"HTTP error code response body: {e.response.text}")
```
nbateshaus: I find the phrasing "HTTP error code response body" confusing; maybe just "HTTP Error body: {e.response.text}"

tamr_unify_client/client.py
```diff
@@ -1,62 +1,76 @@
...
+
 
+# monkey-patch requests.Response.successful
+requests.Response.successful = successful
 
-Response.successful = successful
+
+def _default_response_message(response: requests.Response) -> str:
+    req = response.request
+    return f"{req.method} {response.url} : {response.status_code}"
```
nbateshaus: Does this conform to an existing logging standard?

tamr_unify_client/client.py
```diff
@@ -68,20 +82,13 @@ def __init__(
...
             self.base_path = "/" + self.base_path
 
         if not self.base_path.endswith("/"):
             self.base_path = self.base_path + "/"
 
-        def default_log_entry(method, url, response):
-            return f"{method} {url} : {response.status_code}"
-
-        self.log_entry = None
+        self.response_message = _default_response_message
```
nbateshaus: `_default_response_message` is only used once; should it be in-lined?

nbateshaus: Also, the name `response message` is unclear - this looks quite specific to logging, so `log_message` would be good (or `format_for_log` or similar).
I don't see that it's relevant to put `default` in the name, since there's no way to provide a non-default.

tamr_unify_client/client.py
nbateshaus: Also, note that `requests.response.raise_for_status` can raise URLError (presumably for a Mal-formed URL and the like), and that case is not handled here (which seems fine).

tamr_unify_client/client.py
pcattori: This was a poor-man's way to let users log the response body. I think you're right: we should just inline it and not make it accessible.

tamr_unify_client/client.py
pcattori: Or maybe: `"HTTP Error {e.response.status_code}. Response body: {e.response.text}`?

tamr_unify_client/client.py
pcattori: Agreed, "Check"  could have a boolean connotation. I'll change to "Ensure response does not contain an HTTP error." as the summary description. The "Returns" and "Raises" sections describe the exact behavior.

tamr_unify_client/client.py
pcattori: I don't see how `raise_for_status` directly raises a `URLError` from looking at the [source code](https://requests.readthedocs.io/en/master/_modules/requests/models/#Response.raise_for_status).

tamr_unify_client/client.py
pcattori: Those docs are out-of-date. The stable docs are here: https://requests.readthedocs.io/en/master/

In these [docs](https://requests.readthedocs.io/en/master/api/#requests.Response.raise_for_status) and in the [source code](https://requests.readthedocs.io/en/master/_modules/requests/models/#Response.raise_for_status) for `raise_for_status`, `URLError` is not raised.

tamr_unify_client/client.py
pcattori: Agreed that phrasing needs to be improved. "Checks" can have a boolean connotation. Changing to "Ensure response does not contain an HTTP error.". The "Returns" and "Raises" sections of this docstring cover the exact behavior.

tamr_unify_client/client.py
pcattori: No. I did some digging and found the [format used by `urllib3`](https://github.com/urllib3/urllib3/blob/75d271954b5d662d0567e95b054475e208619dff/src/urllib3/connectionpool.py#L428) (lib used under-the-hood by `requests`).

```python
f'{scheme}://{host}:{port} "{method} {url} HTTP/{http_version}" {status_code} {content_length}'
```

We could crib that....

OR we could just use the logging they already set up instead of writing our own as [shown here](https://2.python-requests.org/en/master/api/#api-changes) (bottom most subsection of "API Changes"). But it looks pretty gnarly...

pcattori: Not only is the setup for `urllib3` logging non-trivial, but you have to set logging level to `DEBUG` to get the request/response info which means you end up getting a lot of cruft:

```sh
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): httpbin.org:443
send: b'GET /headers/jflajsdflkjj HTTP/1.1\r\nHost: httpbin.org\r\nUser-Agent: python-requests/2.22.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nConnection: keep-alive\r\n\r\n'
reply: 'HTTP/1.1 404 NOT FOUND\r\n'
header: Access-Control-Allow-Credentials: true
header: Access-Control-Allow-Origin: *
header: Content-Encoding: gzip
header: Content-Type: text/html
header: Date: Thu, 19 Dec 2019 20:39:31 GMT
header: Referrer-Policy: no-referrer-when-downgrade
header: Server: nginx
header: X-Content-Type-Options: nosniff
header: X-Frame-Options: DENY
header: X-XSS-Protection: 1; mode=block
header: Content-Length: 198
header: Connection: keep-alive
DEBUG:urllib3.connectionpool:https://httpbin.org:443 "GET /headers/jflajsdflkjj HTTP/1.1" 404 198
```

Issue #309, opened 2019-11-01
Dataclasses with type annotations and `from_json` constructors


pcattori:
... instead of forcing all resources to comply with `BaseResource` via inheritance and using `@property`'s to access underlying `_data`.

See [this video](https://www.youtube.com/watch?v=nwjWOaxWMes) for motivation / explanation of `dataclasses` module.
[In this video](https://www.youtube.com/watch?v=T-TwcmT6Rcw), Ramond Hettinger (a Python core contributor) advocates for using `dataclasses`.

Relates to #300 

Example: #303 

## spec

Specs for types defined with `dataclass` should themselves be simple `@dataclass(frozen=True)` that merely provide a `replace` method and a `to_json` method.

pcattori:
Obsoleted by #385

Issue #294, opened 2019-09-17
Create AttributeConfiguration starting with project object


harlankadish:
# 🙋 feature request

I like that AttributeConfiguration.new()... allows me to create an attribute without knowing the underlying JSON. But the only way to use it now is to first use "from tamr_unify_client.project.attribute_configuration.resource import AttributeConfiguration", an approach that I've seen can be unstable.

## 🤔 Expected Behavior

my_project.attribute_cofigurations().new().with_attribute_name("foo").with_similarity_function("GEOSPATIAL_HAUSDORFF")...etc...etc

## 😯 Current Behavior

I need to import the AttributeConfiguration class, ...new().with_<config>(...)put()

## 🔦 Context

I want to provide a script to a customer that (a) doesn't include JSON and (b) is robust to changes in the project-structure of the python client.

PR #330, opened 2020-01-28
Attributes module

pcattori:
# ↪️ Pull Request

Relates to #309 , #300 , #294 

Adds the `attributes` package containing:
- `tc.attribute` and `tc.Attribute`
- `tc.attribute_type` and `tc.AttributeType`
- `tc.subattribute` and `tc.SubAttribute`

Adds the `datasets` package containing:
- `tc.dataset`

The `attributes` and `datasets` packages are type checked via `mypy`.

Additional changes:
- Tests are now being linted by `flake8`
- `invoke` task for typechecking: `typecheck`
- CI performs typechecking
- Reference doc space for features in Beta

## 💻 Examples

```python
import tamr_unify_client as tc

dataset = ...
attr = tc.attribute.create(dataset, name='myattr', is_nullable=False, type=tc.attribute_type.String())

attrs = tc.dataset._attributes(dataset)
assert attr in attrs
```

## ✔️ PR Todo


tamr_unify_client/url.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
+from urllib.parse import urlparse
```
pcattori: remove unused import

tamr_unify_client/dataset/resource.py
```diff
@@ -58,6 +59,11 @@ def attributes(self):
         alias = self.api_path + "/attributes"
         return AttributeCollection(self.client, alias)
 
+    @property
+    def url(self) -> URL:
```
pcattori: need docstring

stubs/responses.pyi
```diff
@@ -0,0 +1,19 @@
...
+def add(
+    method: Optional[str] = None,
+    url: Optional[str] = None,
+    status: Optional[int] = None,
+    json: Optional[JsonDict] = None,
+): ...
+
+T = TypeVar("T")
+
+def activate(T) -> T: ...
```
yizhiyintamr: What does this def activate(T) function do?

pcattori: Context: we are using the `responses` library in our testing suite. This 3rd party library does not come with type annotations, so we have [a couple options](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports) to deal with it:

1. We can simply tell `mypy` to ignore imports from `responses`.
2. We can write our own "stubs" to tell `mypy` what types to expect from the `responses` library.

(1) is easier but it means that `mypy` will let us pass anything to the `responses` library functions and won't complain.

(2) is a little fancier and means we have to write this `.pyi` file, but gives us nice type checking (e.g. `responses.add(asdf=False)` will complain that `asdf` is not a valid keyword to `responses.add` OR `responses.add(method=1)` will complain that `method` should be a `str`, not an `int`).

---

In the `responses` library, the `responses.activate` function is a [decorator](https://dbader.org/blog/python-decorators). Specifically, `responses.activate` takes a function as input and produce a function (with the same signature) as output.

We capture this knowledge here by saying, "for some arbitrary type `T`, the `activate` function should accept something of type `T` and produce something of type `T`".

yizhiyintamr: Decorator article is a very good read. Thanks for the link. It makes sense now.

PR #340, opened 2020-02-25
add version argument to dataset

yizhiyintamr:
# ↪️ Pull Request

Add dataset version argument to dataset for tamr_client

## 💻 Examples

``` python
import tamr_client as tc

auth = tc.UsernamePasswordAuth("admin", "dt")
s = tc.session.from_auth(auth)
instance = tc.Instance(host="10.20.0.186")
dataset = tc.dataset.from_resource_id(s, instance, "49")

print(dataset.version)
```
## ✔️ PR Todo


tamr_client/datasets/dataset.py
```diff
@@ -85,6 +87,7 @@ def _from_json(url: tc.URL, data: JsonDict) -> Dataset:
         name=cp["name"],
         description=cp.get("description"),
         key_attribute_names=tuple(cp["keyAttributeNames"]),
+        version = cp.get("version"),
```
pcattori: I believe that `version` is a required field for datasets, so avoid `.get(...)` since connotes to readers that `version` field might not be present in the JSON (and that `version` property might be set to `None` in the `Dataset`)

```suggestion
        version = cp["version"],
```

PR #339, opened 2020-02-21
TUC: dataset.upsert_from_dataframe

skalish:
# ↪️ Pull Request

This PR adds a convenience function to the dataset object to support upserting records from a pandas DataFrame.  This mirrors the function `datasets.create_from_dataframe` that allows a user to create a dataset from a DataFrame.  Like

Like `create_from_dataframe` the default behavior passes the JSON arg `ignore_nan=True` so that DataFrame nulls are handled gracefully.

Relates to #324 

## 💻 Examples

```python
import pandas as pd
from tamr_unify_client import Client

tamr_client = Client(auth, host)

df_jan = pd.read_csv("USA_Spend_Jan.csv", dtype='str')
primary_key_name = 'contract_transaction_unique_key'
dataset = tamr_client.datasets.create_from_dataframe(df_jan, primary_key_name, 'USA_Spend_Dataset')

df_feb = pd.read_csv("USA_Spend_Feb.csv", dtype='str')
dataset.upsert_from_dataframe(df_feb, primary_key_name)
```

## ✔️ PR Todo

- [x] Added/updated testing for this change
- [x] Included links to related issues/PRs
- [x] Update relevant [docs](https://github.com/Datatamer/tamr-client/tree/master/docs) + docstrings
- [x] Update the [CHANGELOG](https://github.com/Datatamer/tamr-client/blob/master/CHANGELOG.md) under the current `-dev` version:
  - Add changelog entries under any that apply: **BREAKING CHANGES**, **NEW FEATURES**, **BUG FIXES**.
  - Changelog entry format: `[#<issue number>](<link to issue>) <change description>`

Issue #324, opened 2020-01-15
Add convenience function dataset.update_from_dataframe


rhulasi:
# 🙋 feature request

We provide a very useful convenience function to create a dataset from a pandas dataframe.  Would like to be able to have an equivalent function which allows a user to update a dataset in a similar manner

## 🤔 Expected Behavior

Something along the lines of dataset = dataset.update_from_dataframe(df, PK, dataset_name)

## 😯 Current Behavior

Even though the same endpoint is used to create and update records, the rest of the helper functions are not designed to cope with a dataset already existing, so they try to create the dataset from scratch and fail
## 💁 Possible Solution

Workaround:
```
records= df2.to_dict(orient='records')
primary_key_name='ID'
response = dataset.upsert_records(
            records, primary_key_name, ignore_nan=True
)
```
## 🔦 Context
It's great being able to show the python client to partners and those who are less Tamr-savvy but still need to create and update data programatically.  This function could well belong in a library of helper functions, I'll leave that decision up to the dev team :-)

pcattori:
Resolved by #339

PR #348, opened 2020-03-26
TC: Upsert and delete records in dataset + DataFrame support

### LINKED ISSUE #333: Use DataFrame.index to populate primary_key_name in dataset.*.*_from_dataframe
# 🙋 feature request

The method dataset.DatasetCollection.create_from_dataframe() does not play well with Dataframe indexes. If I have a dataframe that is already indexed, I want that index to automatically become my dataset index.

## 🤔 Expected Behavior

Given:
```
>>> records=[{"id":"a","name":"alice bunker","addr":"123 main st"},{"id":"b","name":"archie bunker","addr":"123 main st"}]
>>> df=pd.DataFrame.from_records(data=records, index="id")
>>> df
             name         addr
id                            
a    alice bunker  123 main st
b   archie bunker  123 main st
```

then

```
tamr.datasets.create_from_dataframe(df)
```
should create records with all three attributes ["id", "name", "addr"], and use "id" as the key attribute.

## 😯 Current Behavior

Currently, the index is ignored. If `primary_key_name` is not provided, `create_from_dataframe` fails

## 💁 Possible Solution

```
if primary_key_name is None:
   primary_key_name = df.index.name
```
and corresponding changes to what we feed `upsert_records`

## 🔦 Context

We often use dataframe indexes to aid in data manipulation. It's inconvenient to have to reshape the dataframe to push it to Tamr.

## 💻 Examples

See above.

skalish:
# ↪️ Pull Request

Adds record update and delete functionality to `tamr_client`, similar to that present in `tamr_unify_client`.  Additionally, adds convenience function for updating records from a pandas DataFrame.

Related to #324 #339 
Resolves #333 

## 💻 Examples

```python
import pandas as pd
import tamr_client as tc

instance = tc.Instance(host='10.10.0.121')
dataset = tc.dataset.from_resource_id(s, instance, '1')

records = [{"pk": 1, "val": 11}, {"pk": 2, "val": 12}]

# Upsert these records to dataset
tc.record.upsert(s, dataset, records, primary_key_name="pk")

# Delete these records from dataset
tc.record.delete(s, dataset, records, primary_key_name="pk")

# Upsert records from a DataFrame
df = pd.DataFrame({"pk": [1, 2], "val": [11, 12]})
tc.dataframe.upsert(s, dataset, df, primary_key_name="pk")
```

## ✔️ PR Todo


tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,38 @@
...
+import pandas as pd
+
+import tamr_client as tc
+from tamr_client.datasets.dataset import Dataset
+from tamr_client.datasets.record import PrimaryKeyNotFound
+from tamr_client.types import JsonDict
+
+
+def upsert(
+    session: tc.Session, dataset: Dataset, df: pd.DataFrame, *, primary_key_name: str
```
pcattori: @skalish to support #333 , we could make `primary_key_name` optional and if its omitted use the dataframe's index

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,122 @@
+"""
+See https://docs.tamr.com/reference/record
```
pcattori: As part of this module-level docstring, we could include something like:

"The recommended approach for interacting with records is to use the `upsert` and `delete` functions for all use cases they can handle. For more advanced use cases, the underlying `_update` function can be used directly."

Note: I stole the phrasing from the standard library `subprocess` docs: https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module.

We can also copy this in the docstring for `_update` and replace the current wording to be consistent.

tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,38 @@
+"""
+Convenient functionality for interacting with pandas DataFrames.
+"""
+
+import json
+
+import pandas as pd
+
+import tamr_client as tc
+from tamr_client.datasets.dataset import Dataset
```
pcattori: You can just access `Dataset` via `tc.Dataset`, so you can remove this import. Same could go for `tc.record.PrimaryKeyNotFound`. I think its good practice to use our library internally the same way users would use it, if possible.

Note: I treat `tamr_client.types` as a special case because these emulate types that should exist in the standard library, so I don't like accessing them via `tc.`

tamr_client/__init__.py
```diff
@@ -56,3 +56,9 @@
     AttributeNotFound,
 )
 import tamr_client.attributes.attribute as attribute
+
+# records
+import tamr_client.datasets.record as record
```
pcattori: You could consider additionally `from tamr_client.datasets.record import PrimaryKeyNotFound` so that its available at the top-level as `tc.PrimaryKeyNotFound`. We already do this for other exception types in `tamr_client` above.

tests/tamr_client/datasets/test_record.py
```diff
@@ -0,0 +1,139 @@
...
+@responses.activate
+def test_update():
+    s = utils.session()
+    dataset = utils.dataset()
+
+    url = tc.URL(path="datasets/1:updateRecords")
+    updates = records_to_updates(_records_json)
+    snoop: Dict = {}
+    responses.add_callback(
+        responses.POST, url.__str__(), partial(create_callback, snoop=snoop, status=200)
```
pcattori: The `__str__` method (btw `__<name>__ is pronounced "dunder <name>") should not be called directly. Rather, it allows [the built-in `str` function to do the right thing](https://docs.python.org/3/library/stdtypes.html#str):

```suggestion
        responses.POST, str(url), partial(create_callback, snoop=snoop, status=200)
```

tests/tamr_client/datasets/test_record.py
```diff
@@ -0,0 +1,139 @@
...
+    s = utils.session()
+    dataset = utils.dataset()
+
+    with pytest.raises(tc.record.PrimaryKeyNotFound):
+        tc.record.delete(
+            s, dataset, _records_json, primary_key_name="wrong_primary_key"
+        )
+
+
+def create_callback(request, snoop, status):
```
pcattori: Nice abstraction! Name-wise, what do you think of something like:

```suggestion
def capture_payload(request, snoop, status):
```

or something more descriptive than `create_callback`?

Might be nice to also include a docstring for this function that points back to an example from the `responses` docs

tests/tamr_client/datasets/test_record.py
```diff
@@ -0,0 +1,139 @@
...
+    snoop["payload"] = list(request.body)
+    return status, {}, json.dumps(_response_json)
+
+
+def records_to_deletes(records):
```
pcattori: You should prefix these helper functions with `_` so that its clear they are meant to be local to this file.

tests/tamr_client/datasets/test_record.py
```diff
@@ -0,0 +1,139 @@
...
+        {"action": "CREATE", "recordId": i, "record": record}
+        for i, record in enumerate(records, start=1)
+    ]
+
+
+def stringify(updates):
+    return [json.dumps(u) for u in updates]
+
+
+_dataset_id = "1"
```
pcattori: `_dataset_id`, `_dataset_url`, and `_nan_records_json` don't seem to be used anywhere. Could you remove them?

tamr_client/datasets/dataframe.py
skalish: I'm concerned that using the dataframe index as primary key by default is prone to accidental misuse.  Implementation is pretty simple, but I think at most I could see this as an option that needs to be explicitly set, e.g.
```python
df = pd.DataFrame({"val": [11, 12]})
tc.dataframe.upsert(s, dataset, df, index_as_primary_key=True)
```
where `primary_key` and `index_as_primary_key` are both optional kwargs (default=`None`) with some logic to enforce that exactly one must be passed.

This is definitely a discussion that could benefit from more field input.

tests/tamr_client/datasets/test_record.py
skalish: Agreed on the function name.  Does this seem like something we will want to use in other tests going forward, and if so, should it be located in a different file and then imported?

Also, I'm not sure how to go about pointing to the `responses` doc.

tests/tamr_client/datasets/test_record.py
skalish: Done.

tests/tamr_client/datasets/test_record.py
skalish: Done.

tamr_client/__init__.py
skalish: Done. 👍

tamr_client/datasets/record.py
skalish: I like that phrasing. I don't know if there are any changes I would make to the `_update` docstring.

tamr_client/datasets/dataframe.py
skalish: Done.

tamr_client/__init__.py
skalish: Agreed. Done.

tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+        PrimaryKeyNotFound: If `primary_key_name` is not a column in `df`.
+    """
+    if primary_key_name not in df.columns:
+        raise tc.PrimaryKeyNotFound(
+            f"Primary key: {primary_key_name} is not in DataFrame column names: {df.columns}"
+        )
+    records = json.loads(df.to_json(orient="records"))
```
nbateshaus: this will re-materialize the entire dataframe as a single string, which will definitely be a memory issue, and will almost certainly also be a performance issue

tamr_client/datasets/record.py
pcattori: Can you give `upsert`, `delete` and `_update` ``the :func:`~tamr_client.dataset.upsert_records``` treatment?

tamr_client/datasets/dataframe.py
skalish: I will also bring whatever functionality we settle on here into #339 so the `tuc` DataFrame support will operate in the same way.

tamr_client/datasets/dataframe.py
pcattori: Consider using [`iterrrows`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.iterrows.html) instead of `to_json` to avoid memory issues

skalish: I think we can address that in the function by stringifying and uploading in batches (1000? rows at a time).

A related feature that we've previously discussed is handling the large DataFrame problem with an `upsert_from_csv` function that would cover the issue of not even being able to load everything into a single DataFrame for very large sets of data.

tamr_client/datasets/dataframe.py
pcattori: I don't like including both a `primary_key_name` param and a `index_as_primary_key` param because it allows non-sensical calls like `upsert(..., primary_key_name="non-index column", index_as_primary_key=True)`.

Actually, for upsert, we already have the primary key for the dataset that we are updating via `dataset.key_attribute_names`, so we could make it optional and if omitted, look for that column name.

So this gracefully handles all 3 cases:

```python
# 1. Implicitly use the datasets key attribute name as a primary key
tc.dataset(upsert(s, dataset, df)

# 2. Explicitly name a non-index column as a primary key
tc.dataframe.upsert(s, dataset, df, primary_key_name="non-index pk")

# 3. Opt-in to using the dataframe index as the primary key
tc.dataframe.upsert(s, dataset, df, primary_key_name=df.index.name) 
```

tests/tamr_client/datasets/test_dataframe.py
```diff
@@ -0,0 +1,74 @@
...
+@responses.activate
+def test_upsert():
+    s = utils.session()
+    dataset = utils.dataset()
+
+    url = tc.URL(path="datasets/1:updateRecords")
+    updates = records_to_updates(_records_json)
+    snoop: Dict = {}
+    responses.add_callback(
+        responses.POST, url.__str__(), partial(create_callback, snoop=snoop, status=200)
```
pcattori:
```suggestion
        responses.POST, str(url), partial(create_callback, snoop=snoop, status=200)
```

tests/tamr_client/datasets/test_record.py
pcattori: Let's keep in this file for now, but if we see that we need a similar/same function in other test files, we can *then* factor it out until a more reusable place.

As for docstring linking to `responses` docs, how about:

```python
def _capture_payload(request, snoop, status):
    """Capture request body within `snoop` so we can inspect that the request body is constructed correctly (e.g. for streaming requests).

    See https://github.com/getsentry/responses#dynamic-responses
    """
```

tamr_client/datasets/dataframe.py
skalish: I recall there being some concern that the convenience of case 1 is probably outweighed by the potential for user error.  That is, if there is a column name in `df` that matches the dataset primary key, but does not actually correspond to it, this would result in overwriting specific records unintentionally.

tamr_client/datasets/dataframe.py
pcattori: There are two issues with uploading large datasets:

1. Loading the entire dataset into memory can cause OOM errors.
2. After the client is done sending the data but is still waiting for the response from the server, the server may take too long processing the data. If too much time elapses without data being sent over the connection, the connection will be severed before the server has a chance to send its response.

(1) should be easy to avoid by doing streaming via `iterrows`
(2) might be resolved via chunking, but we should first implement a solution for (1) and then reproduce issue (2) before trying to solve (2)

tests/tamr_client/datasets/test_dataframe.py
skalish: Found and fixed.

tamr_client/datasets/dataframe.py
pcattori: What actually happens if you directly call the API to upsert to a dataset that has an attribute with the same name as the primary key, but uses a separate attribute as the primary key for the upsert? does the upsert-specified PK get used to locate the record but then the PK gets overwritten by the attribute with the same name?

I imagine its pretty common for FEs to name their pks the same thing across creation/upserts, but we could verify this by talking to them.

tests/tamr_client/datasets/test_record.py
pcattori: Hmm, just realized that we do already use this across `test_dataframe` and `test_record` (along with the other helper functions).

We should probably move those to a shared location. `tests.utils` is fine for now

tamr_client/datasets/dataframe.py
skalish: My main motivation for using `.to_json()` is the handling of changing `nan` to `null` now that we aren't using `simplejson`.

I could see incorporating this in the following way:
```python
records = map(lambda x: x[1].to_json(), df.iterrows())
```

I'm interested if there are issues (functionally it seems okay to me, but codestyle seems iffy).

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,124 @@
...
+        records: The records to update, as dictionaries
+        primary_key_name: The primary key for these records, which must be a key in each record dictionary
+
+    Returns:
+        JSON response body from server
+
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+        PrimaryKeyNotFound: If primary_key_name does not match dataset primary key
+        PrimaryKeyNotFound: If primary_key_name not in a record dictionary
```
pcattori: How does specifying `PrimaryKeyNotFound` twice look when the docs are rendered? Could you build the docs locally and see if it comes out ok or if these are deduped?

tamr_client/datasets/dataframe.py
pcattori: Only nit is that for consistency, we should use a generator expression rather than `map`:

```python
records = (x[1].to_json() for x in df.iterrows())
```

tamr_client/datasets/record.py
pcattori: Actually, you also need to add `.rst` files for API reference in `docs/beta` for the relevant functions/exceptions/etc.. in `tc.dataframe` and `tc.record`

tests/tamr_client/datasets/test_record.py
skalish: Should the variables, e.g. `_records_json` stay in test files themselves?

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+    return tc.response.successful(r).json()
+
+
+def upsert(
+    session: tc.Session,
+    dataset: tc.Dataset,
+    records: Iterable[Dict],
+    *,
+    primary_key_name: str,
```
nbateshaus: Why does `primary_key_name` need to be specified? This can be determined by interrogating the dataset.

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,124 @@
...
+    updates = ({"action": "DELETE", "recordId": rid} for rid in record_ids)
+    return _update(session, dataset, updates)
+
+
+def delete(
+    session: tc.Session,
+    dataset: tc.Dataset,
+    records: Iterable[Dict],
+    *,
+    primary_key_name: str,
```
nbateshaus: `primary_key_name` should be fetched from the dataset - asking the user to provide it, when the only right answer is what you get when you interrogate the dataset, is asking for trouble.

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Returns:
+        JSON response body from server
+
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+        PrimaryKeyNotFound: If primary_key_name does not match dataset primary key
+        PrimaryKeyNotFound: If primary_key_name not in a record dictionary
+    """
+    if primary_key_name not in dataset.key_attribute_names:
```
nbateshaus: the API supports datasets with composite keys; tamr-client should support them as well

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,124 @@
...
+        raise PrimaryKeyNotFound(
+            f"Primary key: {primary_key_name} is not in dataset key attribute names: {dataset.key_attribute_names}"
+        )
+    updates = (
+        {"action": "CREATE", "recordId": record[primary_key_name], "record": record}
```
nbateshaus: this will be different for datasets with composite keys

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,124 @@
...
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+        PrimaryKeyNotFound: If primary_key_name does not match dataset primary key
+        PrimaryKeyNotFound: If primary_key_name not in a record dictionary
+    """
+    if primary_key_name not in dataset.key_attribute_names:
+        raise PrimaryKeyNotFound(
+            f"Primary key: {primary_key_name} is not in dataset key attribute names: {dataset.key_attribute_names}"
+        )
+    ids = (record[primary_key_name] for record in records)
```
nbateshaus: this will be different for datasets with composite keys

tamr_client/datasets/record.py
skalish: The motivation here is to ensure that the primary key name in the dataset and in the data match.  If this is not true, the user needs to coerce their data column names to match Tamr.  I was aiming to prioritize avoiding unintentional actions over convenience, since I got the impression from the field that this would be preferred.

tamr_client/datasets/record.py
pcattori: We did a quick spike to investigate composite keys. Apparently, the API accepts composite keys, but the backend can't handle them, so we shouldn't give users the ability to until all parts of the API use composite keys and the backend can actually do meaningful things with composite keys.

tamr_client/datasets/record.py
pcattori: Yea I think the argument is that the data schema *is* part of the interface and doing a column rename to align the primary keys is easy for a user to do.

tamr_client/datasets/dataframe.py
skalish: One issue I've found with this is that each record becomes a JSON-string instead of a dict.  I can handle this by instead using:
```python
records = (json.loads(x[1].to_json()) for x in df.iterrows())
```
which is particularly ugly.
To be clear, what we gain from the above as opposed to 
```python
records = (x[1].to_dict() for x in df.iterrows())
```
is the handling of `np.nan`, without using `simplejson` or writing our own json encoder.

tamr_client/datasets/record.py
skalish: To answer the first question, both PrimaryKeyNotFound entries appear in the docs.

tamr_client/datasets/dataframe.py
pcattori: One option is that you could separate the steps to make a generator pipeline (for readability):

```python
# serialize records via to_json to handle `np.nan` values
serialized_records = (x[1].to_json() for x in df.iterrows())
records = (json.loads(x) for x in serialized_records)
```

tests/tamr_client/datasets/test_record.py
pcattori: Ideally, we could get some data that looks like *real* records rather than just `{"primary_key": 1}` and then we could store the real data as a `.json` (or `.ndjson` for multiple records).

But I'm ok with keeping as is with minor duplication for the data since I could imagine wanting to test these interfaces against different looking data in the future.

tamr_client/datasets/record.py
pcattori: @skalish might be worth at least documenting in the module-level docstring that we don't handle composite keys

tamr_client/datasets/record.py
skalish: Done.

tests/tamr_client/utils.py
```diff
@@ -26,3 +26,29 @@ def dataset():
...
+def capture_payload(request, snoop, status, response_json):
+    """Capture request body within `snoop` so we can inspect that the request body is constructed correctly (e.g. for streaming requests).
+
+    See https://github.com/getsentry/responses#dynamic-responses
+    """
+    snoop["payload"] = list(request.body)
+    return status, {}, json.dumps(response_json)
+
+
+def records_to_deletes(records):
```
pcattori: @skalish what do you think about including the logic to create the DELETE command from a record within `tc.record`:

```python
>>> tc.record._delete_command({"primary_key": 1}, primary_key_name="primary_key")
{"action": "DELETE", "recordId": 1}
```
...and similarly for CREATE command?

Might be overkill, but it seems like a pretty record-specific functionality that doesn't belong in `tests.utils`

skalish: I think that makes sense.  I'll give it a shot.

skalish: @pcattori  Does it make sense to incorporate these into `tc.record.delete()` and `tc.record.create()`?

If so, I think this would deprecate `tc._delete_by_id()` as its only use (unless a user wants to access it directly) is to format the DELETE command.

tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,38 @@
+"""
+Convenient functionality for interacting with pandas DataFrames.
+"""
+
+import json
+
+import pandas as pd
```
pcattori: We need to figure out how to add `pandas` as an optional dependency.

I think we can use [`poetry`'s `extras`](https://python-poetry.org/docs/pyproject/#extras) for this

tests/tamr_client/datasets/test_dataframe.py
```diff
@@ -15,7 +15,10 @@ def test_upsert():
     dataset = utils.dataset()
 
     url = tc.URL(path="datasets/1:updateRecords")
-    updates = utils.records_to_updates(_records_json)
+    updates = [
+        tc.record._create_command(record, primary_key_name="primary_key")
+        for i, record in enumerate(_records_json, start=1)
```
pcattori: `i` is unused, so you no longer need the `enumerate`:

```suggestion
        for record in _records_json
```

Same applies to other usages of `_create_command`, `_delete_command` in the test files.

tamr_client/datasets/record.py
nbateshaus: tamr-client interface should only change in backward-compatible ways to introduce composite key support; right now, primary_key_name is type str - options in the future are to introduce another parameter, e.g. primary_key_names of type list of str, or to make primary_key_name the union of str and list of str

tests/tamr_client/utils.py
```diff
@@ -13,7 +13,7 @@ def load_json(path: Union[str, Path]):
 
 
 def session():
-    auth = tc.UsernamePasswordAuth("admin", "dt")
+    auth = tc.UsernamePasswordAuth("username", "password")
```
pcattori: You should undo this change (i.e. `git rebase -i master` and edit the commit where this change was made so that its not included) and then:

```sh
git checkout master
git pull Datatamer master
git checkout beta-update-dataset
git rebase master
git push -f skalish beta-update-dataset
```
since this change is already on master.

skalish: I think this is fixed now.

tamr_client/datasets/record.py
pcattori: Conversation moved to #349

tamr_client/datasets/dataframe.py
ianbakst: I just tested this behavior. The API will not overwrite the PK, even though I specified the PK field (with a different value) in the record. However, there was no error. The record with the specified recordId was overwritten everywhere I specified except the new PK I had specified.
To put this another way, it seems like every field of the record EXCEPT the PK is eligible to be overwritten.

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,135 @@
...
+        dataset: Dataset containing records to be updated
+        updates: Each update should be formatted as specified in the `Public Docs for Dataset updates <https://docs.tamr.com/reference#modify-a-datasets-records>`_.
+
+    Returns:
+        JSON response body from server
+
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+    """
+    stringified_updates = (json.dumps(update) for update in updates)
```
pcattori, line 37: Maybe try to use `typing.cast` to deal with typechecking here?

```suggestion
    from typing import cast, IO
    ...
    stringified_updates = (json.dumps(update) for update in updates)
    io_updates = cast(IO, stringified_updates)
    ...
    data = io_updates
```

skalish: This fixes the problem.

tamr_client/datasets/dataframe.py
skalish: Moved to #351.

tamr_client/datasets/record.py
```diff
@@ -0,0 +1,137 @@
...
+    # Requests accepts a generator, but typeshed expects this to be a file-like object
```
pcattori, line 38:
```suggestion
    # `requests` accepts a generator, but typing stubs for `requests` in https://github.com/python/typeshed expects this to be a file-like object
```

tamr_client/datasets/dataframe.py
skalish: I think @pcattori's suggested way of handling `primary_key_name` is good.  One consequence that we need to handle is that a DataFrame's index name can be the same as the name of a column, and so this leads to a possible ambiguity.  We could handle this by checking if `primary_key_name` matches both `df.index.name` and a column name in `df.columns` and raise a `ValueError` if there is such a collision.  Incoming commit soon.

tamr_client/datasets/dataframe.py
```diff
@@ -11,28 +12,52 @@
...
+
+    if primary_key_name in df.columns and primary_key_name == df.index.name:
+        raise ValueError(
+            f"Index {primary_key_name} has the same name as column {primary_key_name}"
         )
+
+    if primary_key_name not in df.columns:
+        # if `df.index.name` is the primary key, create column from `df.index`
+        if primary_key_name == df.index.name:
+            df.insert(0, df.index.name, df.index)
```
pcattori, line 46: We should not mutate the passed in dataframe. Instead, we should make a copy and perform mutations on that.

---

Relatedly, the approach here seems to be:
- For datasets with the primary key as a column, do nothing.
- For datasets with the primary key as the index, make the index into a column.

I suggest inverting this approach to:
- For datasets with the primary key as a column, make the column into the index
- For datasets with the primary key as the index, do nothing.

Making the column into an index should be done on a *copy* of the dataframe to avoid mutations to the passed in dataframe.

Then we will have the correct semantics plus we can use *both* values of each item yielded by `iterrows`:

```python
serialized_records = (
    {**row.to_json(), primary_key_name: pk}
    for pk, row in df.iterrows()
) 
```

pcattori: Note: making column into index (and doing this without mutation the original dataframe):

```python
index_df = df.set_index(primary_key_name)
```

Note that `drop` param defaults to `True` and `inplace` defaults to `False` which both work in our favor here.

skalish: That sounds good to me.  I needed to adjust the line setting `serialized_records` a bit, but this seems to work well.

tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Returns:
+        JSON response body from the server
+
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+        PrimaryKeyNotFound: If `primary_key_name` is not a column in `df` or the index of `df`
+        ValueError: If `primary_key_name` matches both a column in `df` and the index of `df`
+    """
+    if not primary_key_name:
```
pcattori, line 35: Technically, `pd.DataFrame` columns can have empty string `''` as a column name, so its better to do an explicit check for `None` (`not primary_key_name` will be `True` for the empty string since empty strings are false-y in Python).

```suggestion
    if primary_key_name is None:
```

tamr_client/datasets/record.py
pcattori: Tracked in #349

tamr_client/datasets/record.py
pcattori: Tracked in #349

tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+    # serialize records via to_json to handle `np.nan` values
+    if primary_key_name == df.index.name:
+        serialized_records = ((pk, row.to_json()) for pk, row in df.iterrows())
+    elif primary_key_name in df.columns:
+        index_df = df.set_index(primary_key_name)
+        serialized_records = ((pk, row.to_json()) for pk, row in index_df.iterrows())
+    else:
+        raise tc.PrimaryKeyNotFound(
```
pcattori, line 50: nit: stylistically, I find it more readable to do all precondition checks with early returns/raises first:

```python
    # preconditions
    if primary_key_name in df.columns and primary_key_name == df.index.name:
        raise ValueError(
            f"Index {primary_key_name} has the same name as column {primary_key_name}"
        )
    elif primary key_name not in df.columns and primary_key_name != df.index.name:
        raise tc.PrimaryKeyNotFound(
            f"Primary key: {primary_key_name} is not DataFrame index name: {df.index.name} or in DataFrame column names: {df.columns}"
        )
```

Then you can handle the serialization as follows:

```python
if primary_key_name in df.columns:
    df = df.set_index(primary_key_name)
serialized_records = ((pk, row.to_json()) for pk, row in df.iterrows())
```

I think this style flows better with the thought process: "make sure the data is formatted correctly. if the primary key is a column, make it an index. ok, now let's stream the upserts."

tamr_client/datasets/dataframe.py
```diff
@@ -0,0 +1,59 @@
...
+    Raises:
+        requests.HTTPError: If an HTTP error is encountered
+        PrimaryKeyNotFound: If `primary_key_name` is not a column in `df` or the index of `df`
+        ValueError: If `primary_key_name` matches both a column in `df` and the index of `df`
+    """
+    if primary_key_name is None:
+        primary_key_name = dataset.key_attribute_names[0]
+
+    if primary_key_name in df.columns and primary_key_name == df.index.name:
+        raise ValueError(
```
pcattori, line 39: Let's go with a custom exception e.g. `AmbiguousPrimaryKey`

Issue #373, opened 2020-04-16
Geospatial equivalent of `create_from_dataframe()`


tiems90:
# 🙋 feature request

Just as normal pandas DataFrames can be created using the `create_from_dataframe()` method,  it would be great if we could do the same for a `GeoPandasDataFrame`. 

## 🤔 Expected Behavior

```
tc.create_from_geodataframe(geodf, primary_key='primary_key', name='my_dataset')
```
Which would create the dataset, generate the attributes—and set the geometry as the correct type. 
## ⚠️ Warning!

When setting the primary key and using the `__geo_interface__` implementation of geopandas, the value in `feature['id']` is based on the index of the GeoDataFrame. 
To manually set this to a different column, `geodf.index = geodf['my_primary_key']` can be used. 

## 🔦 Context

Using the geoJSON API endpoint is less than ideal as all column fields need to be specified in the curl command. 
Having it integrated neatly would also allow us to leverage the many types of geospatial data that geopandas can read. (shape files, etc.)

Issue #351, opened 2020-04-01
Handle pandas as an extra dependency


skalish:
# 🙋 feature request
There are some external packages that are required for convenience features desired in the Tamr client, e.g. `pandas` for DataFrame support, that are difficult or impossible to have installed in some environments.  These dependencies can be changed from required or dev-dependencies into optional extras: https://python-poetry.org/docs/pyproject/#extras

Related to #324, #348 

## 💁 Possible Solution
Move `pandas` from `[tool.poetry.dev-dependencies]` to `[tool.poetry.extras]` in `pyproject.toml`.  In the related code that imports `pandas` this import could be wrapped in a `try`-`except` statement that raises a to-be-defined exception if the extra dependency has not been installed.

## 💻 Examples
```bash
poetry install --extras "pandas"
```

pcattori:
@skalish note that the extras name should semantically be a *feature* name, not a dependency name. So "dataframe" (after the module name) might be more apt than "pandas"

pcattori:
Resolved by #398

PR #375, opened 2020-04-20
Add create_from_geodataframe functionality

tiems90:
# ↪️ Pull Request
Linked to Feature Request: #373 
Provides a convenience function that creates a dataset, adds attributes as needed, and upserts records to tamr. 

## Feedback wanted:
- Converting GDF columns to strings is not as easy as calling `df.astype('str')` in pandas, since changing the geometry column type converts it back to a normal pandas df. The conversion could be done automatically, but it might be better to be consistent with #320.
- I've set a default geometry attribute name for now, in line with the common naming from `geopandas`. Yay/nay?
- Putting the entire geospatial spec in the body doesn't look very nice. Would it be worth factoring it out? 

## 💻 Examples

```
gdf = gpd.read_file("/test.geojson")
gdf[gdf.columns.difference(['geometry'])] = gdf[gdf.columns.difference(['geometry'])].astype('str')

dataset = tamr.datasets.create_from_geodataframe(gdf, 'primaryKey', 'test')
```

## ✔️ PR Todo

- [ ] Added/updated testing for this change:
  - Did local testing, but I am not familiar with the unit tests for API calls, so could use help there. 


tamr_unify_client/dataset/collection.py
```diff
@@ -171,6 +171,189 @@ def create_from_dataframe(
...
+
+        Each attribute has the default type `ARRAY[STRING]`, besides the key attribute, which will have type `STRING`,
+        and the geometry attribute `geo_attr` which will have the type `GEOSPATIAL`.
+
+        This function attempts to ensure atomicity, but it is not guaranteed. If an error occurs while creating
+        attributes or records, an attempt will be made to delete the dataset that was created. However, if this
+        request errors, it will not try again.
+
+        :param geodf: The data to create the dataset with.
+        :type df: :class:`geopandas.geodataframe.GeoDataFrame`
```
pcattori, line 188: Type information should be encoded as type annotations (see #300 , for an example see https://github.com/Datatamer/tamr-client/blob/master/tamr_unify_client/dataset/resource.py#L89 )

tamr_unify_client/dataset/collection.py
```diff
@@ -171,6 +171,189 @@ def create_from_dataframe(
...
+        (with `primary_key_name` as the key attribute), and upserts a record for each row of `geodf`.
+
+        Each attribute has the default type `ARRAY[STRING]`, besides the key attribute, which will have type `STRING`,
+        and the geometry attribute `geo_attr` which will have the type `GEOSPATIAL`.
+
+        This function attempts to ensure atomicity, but it is not guaranteed. If an error occurs while creating
+        attributes or records, an attempt will be made to delete the dataset that was created. However, if this
+        request errors, it will not try again.
+
+        :param geodf: The data to create the dataset with.
```
pcattori, line 187: Docstrings should be written in Google-style (see #300 )

tamr_unify_client/dataset/collection.py
```diff
@@ -171,6 +171,189 @@ def create_from_dataframe(
...
+            if col == primary_key_name:
+                # this attribute already exists, so don't create it again
+                continue
+
+            if col == geo_attr:
+                # this column has a different creation spec
+                attr_spec = {
+                    "name": geo_attr,
+                    "description": "",
+                    "type": {
```
pcattori, line 222: Let's take the `type` out and put it into a proper JSON file and just read it in here, rather than inlining the full type since it's so large.

tamr_unify_client/dataset/collection.py
```diff
@@ -171,6 +171,189 @@ def create_from_dataframe(
...
+            raise CreationError("Dataset was not created")
+        # after this point, if a request fails, try to undo the change by deleting this dataset
+
+        attributes = dataset.attributes
+        for col in geodf.columns:
+            if col == primary_key_name:
+                # this attribute already exists, so don't create it again
+                continue
+
+            if col == geo_attr:
```
pcattori, line 217: Can there be more than one geo attribute in a Geo DataFrame? If so, we should account for that here. Tamr is certainly able to handle datasets with multiple geo attributes.

Also is there a way to detect Geo attributes by their `dtype` or similar in GeoPandas so we can always do the right thing instead of relying on users to pass in the correct names?

tamr_unify_client/dataset/collection.py
tiems90: Yep, for now I went with consistency of `create_from_dataframe`, but obviously just reading the type from somewhere else would be much nicer.

tamr_unify_client/dataset/collection.py
tiems90: Yes, it can have more than one geo_attribute (encoded as `geoSeries`) in one geoDF, but only one will be the 'active' in terms of which one is called when calling methods on the geoDF. 

Default will be `geometry` when reading in most file types in geopandas, but if people have multiple types, it should be set explicitly.

tamr_unify_client/dataset/collection.py
tiems90: In order to specify the type `geodf: gpd.GeoDataFrame` I would need to import geopandas as in the [example](https://github.com/Datatamer/tamr-client/blob/6924c26975e66e58933ee75acde48fa68e2bc09a/tamr_unify_client/dataset/resource.py#L3)

Are we happy to do that? It would create a dependency on geopandas.

tamr_unify_client/dataset/collection.py
tiems90: Updated to google docstring format

tamr_unify_client/dataset/collection.py
tiems90: Causes checks to fail since the package is not  installed in pytest.

tamr_unify_client/dataset/collection.py
tiems90: I've now included a detection for geometry columns, which will be passed a spec of geometry type.

tamr_unify_client/dataset/collection.py
tiems90: Included:

`from tamr_unify_client.attribute.specs.tamr_geometry_type import TAMR_GEOMETRY_TYPE`

tamr_unify_client/dataset/collection.py
pcattori: I think a dependency on GeoPandas is ok here as long as there aren't any gotchas for installation. @tiems90 is GeoPandas easy to install? (just a `pip install` or does it require other stuff?)

In the `tamr_client` BETA, we have been packaging things by *feature* and have been planning to include optional features (see #351 ) so we could include GeoPandas as an optional feature there if needed. But I think its worth straight up including as a dependency first. If people have issues with GeoPandas as a dependency, they can file issues on this repo and then we can make into an optional dependency.

tamr_unify_client/dataset/collection.py
pcattori: Instead of `attribute/spec/tamr_geometry_type.py`, could you encode this in an actual JSON file (not a Python file with a `dict` that corresponds to that JSON) and read in via `json.load(f)`?

We can structure as `attribute/types/tamr_geometry.json`.

tamr_unify_client/dataset/collection.py
tiems90: Ok! A pip install should do.
Some of the functionality requires additional libraries installed `apt-get install` style, though.

tamr_unify_client/dataset/collection.py
pcattori: Also, only the actual `type` should be encoded in the JSON file, so the resulting code would look like:

```python
import json
# pretend we have `tuc` defined as a path pointing to `tamr_unify_client/` dir
with open(tuc / "attribute" / "types" / "tamr_geometry.json") as f:
    type = json.load(f)

for attr_name in geo_attr_names:
    attr = {
        "name": attr_name,
        "description": "",
        "type": type,
        ...
    }
```

tiems90: Sure, did it to allow easy importing and relative paths. 
Does poetry include a `pkg_resources` equivalent?

pcattori: Yes, see https://python-poetry.org/docs/pyproject/#include-and-exclude

tamr_unify_client/dataset/collection.py
pcattori: Hmm, this is where I am out of my depth regarding geospatial stuff. @nbateshaus is limitation on having at most one geo attribute something we impose in `tamr-client` or something imposed by our underlying dependencies?

In either, case what do you recommend?

tiems90: It's possible to convert the geom that isn't the primary one to a string and upsert to tamr; not sure if that's a solution.

tamr_unify_client/dataset/collection.py
tiems90: And we're back to why I detest the import functionality of Python. For now *real* work calls so I'll look at.

tamr_unify_client/dataset/collection.py
tiems90: Added `geopandas` to poetry. It now passes the checks.

pcattori:
@tiems90 did you need to explicitly add the `.json` file to the `pyproject.toml` under `include`?

One way to test this is to try to build the project (`poetry build`) and then create a dummy virtualenv somewhere and try to `pip install <path/to/tamr-unify-client/dist/something.whl>` and see if your functionality works there.

tiems90:
@pcattori Nope! Did not have to be declared.

tamr_unify_client/dataset/collection.py
tiems90: @nbateshaus Any thoughts on how you would want this to work with the `tamr-client`. For now it uses `dataset.from_geo_features()` but this does not allow for upserting multiple geometries?

The way I see it we can:
[ ] Implement this now for only one geometry to be used, ignoring others.
[ ] Implement this now for only one geometry to be used, upserting the others as string
[ ] Implement multiple geometries in `dataset.from_geo_features()` and then use that here. 

The last one seems difficult as I understand it uses the `__geo_interface__`.

nbateshaus: `__geo_interface__` is restricted to a single geometry, so we should not use that.
We should use logic similar to `from_geo_features()` to get all the geometries into the geodataframe.

More broadly, `__geo_interface__` is a manifestation of Simple Feature Access which says that a given "feature" can have at most one geometry. That is only one specific standard for working with geospatial information.

Geodataframe (and Tamr Tamr) are less restrictive. Just as we built interoperability between Tamr dataset and `__geo_interface__`, we should build interoperability between Tamr dataset and geodataframe - but this will have to be a completely separate implementation, separate from anything `__geo_interface__` related.

tiems90: @nbateshaus 
I agree on the interoperability; this wrapper creates multiple geospatial attributes in a new `dataset` based on the type within geopandas. My issue is with upserting/serialising these. 

I run into issues with `from_geo_features()`. Although the docs mention it is supposed to work with multiple geometries, I understand it only upserts one geometry column feature per record (given the underlying call to `_feature_to_record()` which takes `geo_attr: The **singluar** attribute on the record to use for the geometry`.)

So the current implementation of `from_geo_features()` appears to ignore all other geometry attributes in the GDF. This implementation carries this forward. The question is more whether we want to upsert multiple geometries per record; which would seem to require more changes. 

There's also the very probable case that I've missed something essential, in which case an example of how to upsert multiple geometries per records would be great so I can make the changes 😊

tiems90:
Since AFAIK there's no internal functions to wrap around to deal with multiple geometries I'll go ahead and implement this outside of the client.

PR #381, opened 2020-04-30
Refactored Namespacing for datasets

ianbakst:
Refactored namespace for datasets module in BETA. Dataset pulled up to tc.datasets level in tamr_client/datasets/__init__.py and then further in tamr_client/__init__.py. Furthermore, DatasetNotFound exception is now just NotFound and only pulled up to datasets level, and will read tc.dataset.NotFound when raised.

💻 Examples
✔️ PR Todo
[] Added/updated testing for this change
[] Included links to related issues/PRs
[] Update relevant docs + docstrings
[] Update the CHANGELOG under the current -dev version:
Add changelog entries under any that apply: BREAKING CHANGES, NEW FEATURES, BUG FIXES.
Changelog entry format: [#<issue number>](<link to issue>) <change description>


tamr_client/datasets/__init__.py
```diff
@@ -1,2 +1,5 @@
 # This __init__.py file is necessary for `mypy --package`
 # See https://github.com/python/mypy/issues/5759
+
+from tamr_client.datasets import Dataset
+import tamr_client.datsets
```
pcattori, line 5: typo:

```suggestion
import tamr_client.datasets
```

tamr_client/datasets/__init__.py
```diff
@@ -1,2 +1,5 @@
 # This __init__.py file is necessary for `mypy --package`
```
pcattori, line 1: we can remove this comment since it was just here to justify the presence of an empty `__init__.py` file, but it's not empty anymore.

tamr_client/__init__.py
```diff
@@ -37,8 +37,8 @@
 import tamr_client.session as session
 
 # datasets
-from tamr_client.datasets.dataset import Dataset, DatasetNotFound
-import tamr_client.datasets.dataset as dataset
+from tamr_client.datasets import Dataset
```
pcattori, line 40: Let's also rename the package from `tc.datasets` to just `tc.dataset`.

ianbakst: so would it just be `import tamr_client.dataset` or `import tamr_client.dataset as dataset`?

pcattori: We want users to be able to do:

```python
import tamr_client as tc # single import for the whole library

x = tc.dataset.<some dataset function>(...)
```

`import tamr_client as tc` will make everything that is available in `tamr_client/__init__.py` available *by the same name* it has within `__init__.py`. 

So if you just do `import tamr_client.dataset` within `tamr_client/__init__.py`, then the name `dataset` is *not* available within `tamr_client/__init__.py` (you would still have to reference as `tamr_client.dataset` within `tamr_client/__init__.py`).

That means that the user would have to do:

```python
import tamr_client as tc # single import for the whole library

x = tc.tamr_client.dataset.<some dataset function>(...)
```

So you need to do `import tamr_client.dataset as dataset` so that the name `dataset` is available in `tamr_client/__init__.py` so that you can do `tc.dataset.<some dataset function>(...)`.

ianbakst:
Here's a hack at something a little different. This should probably also be a draft PR now.
Also, I meant "refactor" not "rebase".

tamr_client/dataset/__init__.py
```diff
@@ -0,0 +1,11 @@
+# See https://github.com/python/mypy/issues/5759
```
pcattori, line 1: You'll want to start this file with:

```python
# flake8: noqa
```

to turn off linting for this file (since it will complain about importing a bunch of things that aren't used within this file)

See https://github.com/Datatamer/tamr-client/blob/master/tamr_client/__init__.py#L1

tamr_client/dataset/dataframe.py
```diff
@@ -7,9 +7,11 @@
 
 import pandas as pd
 
-import tamr_client as tc
 from tamr_client.types import JsonDict
-
+from tamr_client.dataset import Dataset
```
pcattori, line 11: To keep internal `import`s from depending on `__init__.py` hoisting, this should be:

```suggestion
from tamr_client.dataset.dataset import Dataset
```

(Note: this repetition is fine. we can do a follow-on PR where we split up the data class and functions and rename modules where the `dataset.dataset` repetition will be addressed)

tamr_client/dataset/dataframe.py
```diff
@@ -8,7 +8,7 @@
 import pandas as pd
 
 from tamr_client.types import JsonDict
-from tamr_client.dataset import Dataset
+from tamr_client.dataset.dataset import Dataset
 from tamr_client.session import Session
 from tamr_client.dataset import dataset
```
pcattori, line 13: This is importing the `dataset.py` module from the `tamr_client.dataset` package via import hoisting in `tamr_client/dataset/__init__.py`.

To import the `dataset.py` module without using import hoisting:

```suggestion
import tamr_client.dataset.dataset as dataset
```

Same goes for the `record` import below this one and other imports in other files in this PR.

pcattori:
Closing in favor of doing a prefactoring PR for converting all internal imports to absolute imports (that import a module or value from a module, *not* a package) and a separate PR for the dataset package refactor.

PR #382, opened 2020-05-06
Changed imports to be absolute everywhere.

ianbakst:
# ↪️ Pull Request

Changed the imports in all the modules to absolute paths. Also took steps to resolve circular dependence between SubAttribute and attribute_type
## 💻 Examples

`import tamr_client as tc` is no longer in the modules, instead the modules/objects required by a given module are directly imported e.g. `from tamr_client.datasets.dataset import Dataset`

## ✔️ PR Todo

### DIFF
--- a/tamr_client/attributes/attribute.py
+++ b/tamr_client/attributes/attribute.py
@@ -5,7 +5,13 @@ from copy import deepcopy
 from dataclasses import dataclass, field, replace
 from typing import Optional, Tuple
 
-import tamr_client as tc
+from tamr_client.url import URL
+from tamr_client.attributes.attribute_type import AttributeType
+from tamr_client.attributes import attribute_type
+from tamr_client.attributes import type_alias
+from tamr_client.session import Session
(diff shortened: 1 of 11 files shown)


skalish:
I think this looks good. Can we explicitly the development style that we are establishing here? From what I can tell, specifically for writing imports in `tamr-client` importing from `tamr-client`:
- All such imports should be absolute imports to the actual location of the object
- When a function from a sub-package is needed, the sub-package (and not the function itself) should be imported and aliased as the subpackage name (dropping the full path)
  - YES: `import tamr_client.response as response` then `response.successful(r)`
  - NO: `import tamr_client.response.successful as successful` then `successful(r)`
  - NO: `import tamr_client.response` then `tamr_client.response.successful(r)`
- Dataclass/type definitions should be imported directly
  - YES: `from tamr_client.datasets.dataset import Dataset`
  - NO: `import tamr_client.datasets.dataset as dataset` then refer to `dataset.Dataset`

Am I being incorrect/too strict/not strict enough with this statement?

PR #384, opened 2020-05-15
Rename dataset

ianbakst:
# ↪️ Pull Request
Renamed the `datasets` module to `dataset` to avoid further confusion.

## ✔️ PR Todo


skalish:
The failed `Docs` task seems to be related to how the docs are being generated now that there is ambiguity between `tc.dataset` and `tc.dataset.dataset`.  I was able to generate them successfully by changing a line in `docs/beta/datasets/dataset.rst` from
```python
.. autofunction:: tamr_client.dataset.from_resource_id
```
to
``` python
.. autofunction:: tamr_client.dataset.dataset.from_resource_id
```
This doesn't seem like the right solution though, since the intended method of accessing this function is the former, and the docs generated from this suggest otherwise.

pcattori:
@skalish @ianbakst : the way I'd suggest for fixing the ambiguity between `tc.dataset` and `tc.dataset.dataset` is to rename the `dataset.py` module something else.

One obvious choice is `_dataset.py`. This is a common practice in other Python codebases where imports, package names and module names conflict (e.g. pandas codebase has many modules/packages named with a `_` prefix).

skalish:
@pcattori It turns out the ambiguity wasn't causing the doc creation problem; instead it's that we are not yet hoisting the function `from_resource_id` in `datasets/__init__.py`.

I think this is an  argument for including the `dataset`-related import changes in this PR.  As far as I can tell, if we don't do this now, either the docs fail to build, or the docs will refer to this function as `tamr_client.dataset.dataset.from_resource_id(...)` which is not the way it should be accessed by a user.

@ianbakst, please chime in if it turns out we were wrong about this when you make the changes.

pcattori:
You'll need a `# flake8: noqa` comment at the top of non-empty `__init__.py` files like https://github.com/Datatamer/tamr-client/blob/master/tamr_client/__init__.py#L1

tamr_client/dataset/__init__.py
```diff
@@ -0,0 +1,8 @@
+# flake8: noqa
+# See https://github.com/python/mypy/issues/5759
```
pcattori, line 2: This comment is only relevant for empty `__init__.py` files.

```suggestion
```

tamr_client/dataset/__init__.py
```diff
@@ -0,0 +1,7 @@
+# flake8: noqa
+from tamr_client.dataset.dataset import Dataset
+
+from tamr_client.dataset.dataset import from_resource_id
+from tamr_client.dataset.dataset import DatasetNotFound
```
pcattori, line 5: Could you combine imports from the same module:

```python
from tamr_client.dataset.dataset import DatasetNotFound, from_resource_id
```

pcattori, line 5: `DatasetNotFound` should also be renamed (here and in `dataset.py`) to make sure of namespacing:

```suggestion
from tamr_client.dataset.dataset import NotFound
```

You'll likely need to change this in a couple tests and in the docs too.

tamr_client/dataset/dataframe.py
```diff
@@ -7,8 +7,8 @@
 
 import pandas as pd
 
-from tamr_client.datasets.dataset import Dataset
-import tamr_client.datasets.record as record
+from tamr_client.dataset.dataset import Dataset
+import tamr_client.dataset.record as record
```
skalish, line 11:
```suggestion
from tamr_client.dataset import record
```

PR #388, opened 2020-06-09
TC: Stream records from a dataset

skalish:
# ↪️ Pull Request

This PR adds record streaming functionality to `tamr_client` through the function `tamr_client.datasets.record.stream`.  It is modeled after the existing method in TUC.

Related to #11 

## 💻 Examples

```
import tamr_client as tc

dataset = ...

records = tc.record.stream(s, dataset)  # this is a generator that returns records as dicts
```

## ✔️ PR Todo


skalish:
1. This should probably have more testing.  Not sure if the existing test should be written differently.

2. Should this be a part of the module `record.py` or `dataset.py`?

tamr_client/dataset/record.py
```diff
@@ -143,3 +143,17 @@ def _delete_command(record: Dict, *, primary_key_name: str) -> Dict:
...
+def stream(session: Session, dataset: Dataset) -> Generator[Dict, None, None]:
+    """Stream the records in this dataset as Python dictionaries.
+
+    Args:
+        dataset: Dataset from which to stream records
+
+    Returns:
+        Python generator yielding records
+    """
+    with session.get(str(dataset.url) + "/records", stream=True) as response:
```
pcattori, line 157: You should use [tc.response.ndjson](https://github.com/Datatamer/tamr-client/blob/c08704de48eacb0d8f992ca9dfca80df17018d5b/tamr_client/response.py#L34) for this instead.

tamr_client/dataset/record.py
```diff
@@ -143,3 +143,17 @@ def _delete_command(record: Dict, *, primary_key_name: str) -> Dict:
         The DELETE command in the proper format
     """
     return {"action": "DELETE", "recordId": record[primary_key_name]}
+
+
+def stream(session: Session, dataset: Dataset) -> Generator[Dict, None, None]:
```
pcattori, line 148: You should use `Iterator` type for generators that do not read from `yield` nor `return`. see https://docs.python.org/3/library/typing.html#typing.Generator

Also, you can use `types.JsonDict` instead of `Dict`

```suggestion
def stream(session: Session, dataset: Dataset) -> Iterator[JsonDict]:
```

tests/tamr_client/datasets/test_record.py
```diff
@@ -154,6 +155,20 @@ def test_delete_infer_primary_key():
...
 
 
+@responses.activate
+def test_stream():
+    s = utils.session()
+    dataset = utils.dataset()
+
+    url = tc.URL(path="datasets/1/records")
+    responses.add(
+        responses.GET, str(url), body="\n".join([json.dumps(x) for x in _records_json])
```
pcattori, line 165: You should use a generator expression instead of list comprehension to avoid unnecessary `[]` and to avoid loading the data into memory separately from creating the `str`:

```suggestion
        responses.GET, str(url), body="\n".join(json.dumps(x) for x in _records_json)
```

tests/tamr_client/datasets/test_record.py
```diff
@@ -154,6 +155,20 @@ def test_delete_infer_primary_key():
...
+    )
+
+    response = tc.record.stream(s, dataset)
```
pcattori, line 168: nit: `response` connotes that this object is an HTTP response. You could just use `records` instead:

```suggestion
    records = tc.record.stream(s, dataset)
```

pcattori:
@skalish I think the testing is sufficient for now since its at least on par with [TUC testing for record streaming](https://github.com/Datatamer/tamr-client/blob/master/tests/unit/test_dataset_records.py#L19)

I also think `tc.record.stream` is better than `tc.dataset.stream` since that would probably be too vague and need to be changed to `tc.dataset.stream_records` at which point you are already aware of records anyway. Additionally, the vAPI models records as a subresource of the dataset (not as part of the dataset resource itself), so this is consistent with that.

Issue #372, opened 2020-04-16
TC: Design for unified datasets


pcattori:
# 💬 RFC

Now that we have projects in `tamr_client`, the next logical step is to add support for unified datasets.

Proposal: add a new module, `tc.datasets.unified`, with unified dataset features:
- get unified dataset from project
- get name of unified dataset
- commit unified dataset
- ...more?

## 🔦 Context

I write a lot of functions that accept a mastering project as input and then need to cobble together names of derived datasets based on the unified dataset name.

## 💻 Examples

using `tamr_unify_client`:

```python
def _compile_metrics(project: MasteringProject) -> Metrics:
    ud = project.unified_dataset()

    # keep in-memory lookup of group -> true cluster ID
    grouped_entities = project.client.datasets.by_name(
        f"{ud.name}_dedup_grouped_entities"
    )
    ...
```

using `tamr_client`:

```python
def _compile_metrics(s: tc.Session, project: tc.mastering.Project) -> Metrics:
    ud = tc.datasets.unified.from_project(s, project)

    # keep in-memory lookup of group -> true cluster ID
    instance = project.url.instance
    grouped_entities = tc.datasets.from_name(
        s,
        instance, 
        f"{ud.name}_dedup_grouped_entities"
    )
    ...
```
```

nbateshaus:
A challenge for users of this structure is that the association between projects and unified datasets has to live in the user's head; if a user is looking at a project, they have no idea that a unified dataset is a thing that might be associated with it.

nbateshaus:
There are many derived datasets; a unified dataset is one example. Derived dataset have a lot in common - in particular, they cannot be updated, their schema cannot be modified, and they can be refreshed. I would prefer if a derived dataset were a thing, and a unified dataset were a way to identify a particular derived dataset associated with a project.

ianbakst:
I am inclined to agree with the refined structure that @nbateshaus just described. That might be a good way to group datasets. Would it make sense for the user could be able to get a Unified Dataset either: 1- from a project (if they do understand the project-UD relationship), or 2- by_name (if they don't know that relationship or don't care about any other aspects of the project)? Either one would return the UD as a derived dataset. ...I might be missing the point, though.

ianbakst:
At least from my (FE) perspective, the duality of a dataset, especially a derived dataset, as being both a standalone item and associated with (or even a direct output of) a project is very real in my head.

pcattori:
[...]

PR #389, opened 2020-06-13
Unified Dataset dataclass

ianbakst:
# ↪️ Pull Request

Added the Unified Dataset dataclass. It represents a unifiedDataset. The Unified Dataset and (regular) Dataset objects are brought together in a Union dataclass. There are two major functions in this module: from_project, which returns a unified dataset from its project, and commit, which commits the unified dataset (materializes).
#372 
## 💻 Examples

if you have a project object: project
`unified_dataset = tc.unified.from_project(session, instance, project)`
you can then commit that unified dataset by:
`tc.unified.commit(unified_dataset,session)`

## ✔️ PR Todo


tamr_client/dataset/dataset.py
```diff
@@ -37,6 +38,9 @@ class Dataset:
     description: Optional[str] = None
 
 
+AllDataset = Union[Dataset, UnifiedDataset]
```
pcattori, line 41: To be consistent with the vernacular of the built-in `typing` module, we should name the union "AnyDataset":

```suggestion
AnyDataset = Union[Dataset, UnifiedDataset]
```

tamr_client/dataset/unified.py
```diff
@@ -0,0 +1,106 @@
...
+    cp = deepcopy(data)
+    return Dataset(
+        url,
+        name=cp["name"],
+        description=cp.get("description"),
+        key_attribute_names=tuple(cp["keyAttributeNames"]),
+    )
+
+
+def commit(unified_dataset: Dataset, session: Session) -> JsonDict:
```
pcattori, line 95: The convention in `tamr_client` is to always have `session` as the first arg if its present:

```suggestion
def commit(session: Session, unified_dataset: Dataset) -> JsonDict:
```

tamr_client/dataset/unified.py
```diff
@@ -0,0 +1,106 @@
...
+
+from tamr_client import response
+from tamr_client.instance import Instance
+from tamr_client.project import Project
+from tamr_client.session import Session
+from tamr_client.types import JsonDict
+from tamr_client.url import URL
+
+
+class DatasetNotFound(Exception):
```
pcattori, line 16: Already in the `dataset/unified` namespace, so this should just be named "NotFound":

```suggestion
class NotFound(Exception):
```

CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 ## 0.12.0-dev
   **BETA**
-
   Important: Do not use BETA features for production workflows.
-  
+  - [#372](https://github.com/Datatamer/tamr-client/issues/372) TC:Design for unified datasets
```
skalish, line 4: I think this should go at the bottom of this section, below 
```
- Support for streaming records from a dataset via `tc.record.stream`
```
so that these are in chronological order.

skalish:
Now that we have `AnyDataset` defined, functions that currently take `Dataset` as an argument should be updated to take `AnyDataset` **if** this is valid.  For instance, `dataset.record.stream` should be changed (but we should make sure it does work using the project url), but `dataset.record.upsert` should not be changed since you can't modify records in a unified dataset.

CHANGELOG.md
ianbakst: It was my assumption that they should be in chronological order with most recent at top.

pcattori: Historically, we haven't cared about order for changes that are part of the same release since those changes become user-facing at the same time.

skalish:
> `attribute.py` also has functions that act on datasets, and it looks like there are relevant versioned API calls for these on unified datasets.
After some offline discussion, it sounds like these should be handled carefully, probably in a future PR that further fleshes out attribute operations for datasets and projects.

tests/tamr_client/dataset/test_unified_dataset.py
```diff
@@ -0,0 +1,53 @@
+import pytest
```
pcattori, line 1: The name of this file should mirror the file name in the source code, so should be changed to `test_unified.py`

tamr_client/dataset/unified.py
```diff
@@ -0,0 +1,106 @@
...
+class NotFound(Exception):
+    """Raised when referencing (e.g. updating or deleting) a unified dataset
+    that does not exist on the server.
+    """
+
+    pass
+
+
+@dataclass(frozen=True)
+class Dataset:
```
pcattori, line 25: We've been making class names that are user-facing globally unique to make them easily distinguishable (e.g. if you print the `repr` for an instance of any class). This is a known exception to the "use namespaces instead of redundancy in names" rule.

```suggestion
class UnifiedDataset:
```

If we do this, then we should make sure to hoist the `UnifiedDataset` type so its available at the top-level as `tc.UnifiedDataset`.

We should probably discuss this during OH.

If we continue using this convention, we should probably codify it in the style guide section of the contributor docs.

PR #390, opened 2020-06-17
TC: Add operations

skalish:
# ↪️ Pull Request

This PR adds operations to the Tamr Client beta, largely mirroring this feature in TUC.

## 💻 Examples
```python
session = ...
instance = ...
unified_dataset = ...
response = tc.dataset.unified.commit(session, unified_dataset) # not quite accurate, see * below
operation = tc.operation.from_response(instance, response) 

print(tc.operation.succeeded(op))  # check if operation had succeeded when last polled

op = tc.operation.poll(session, op)  # update operation from the server

tc.operation.wait(session, op)  # wait until operation stops running (succeeded/failed/canceled)
```
*Really, once this is implemented, `tc.unified.commit()` should return an `Operation` instance

## ✔️ PR Todo


skalish:
This PR still needs confirmation on structure and possibly more test coverage.

tamr_client/operation.py
```diff
@@ -0,0 +1,195 @@
...
+            "relativeId": "operations/-1",
+        }
+    else:
+        resource_json = response.json()
+    _id = resource_json["id"]
+    _url = URL(instance=instance, path=f"operations/{_id}")
+    return _from_json(_url, resource_json)
+
+
+def apply_options(
```
pcattori, line 82: same comment about `_` prefix:

```suggestion
def _apply_options(
```

This would let us remove the `**NOTE**` in the docstring too.

tamr_client/operation.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@dataclass(frozen=True)
+class Operation:
+    """A Tamr operation
+
+    See https://docs.tamr.com/new/reference/the-operation-object
+
+    Args:
+        url
```
pcattori, line 32: Are the other attributes intentionally left out?

```suggestion
        url
        id
        type
        status
        description
```

tamr_client/operation.py
pcattori: There's a design question here about whether to include this function at all, or if users should just call `wait` themselves explicitly.

Additionally, it would be useful to ask the field how often, if ever, they use `asynchronous` mode.

tamr_client/operation.py
```diff
@@ -0,0 +1,195 @@
...
+    return Operation(
+        url,
+        id=cp["id"],
+        type=cp["type"],
+        status=cp.get("status"),
+        description=cp.get("description"),
+    )
+
+
+def poll(session: Session, operation: Operation) -> Operation:
```
pcattori, line 147: nit: move the intended interface functions to the top, and the `_`-prefixed functions to the bottom of this file

tamr_client/operation.py
```diff
@@ -0,0 +1,195 @@
...
+        timeout_seconds: Time (in seconds) to wait for operation to resolve.
+
+    Raises:
+        TimeoutError: If operation takes longer than `timeout_seconds` to resolve.
+    """
+    started = now()
+    while timeout_seconds is None or now() - started < timeout_seconds:
+        if operation.status and operation.status["state"] in ["PENDING", "RUNNING"]:
+            sleep(poll_interval_seconds)
+        elif not operation.status or operation.status["state"] in [
```
pcattori, line 180: we can probably simplify the logic here by using a guard clause for `operation.status` first:

```python
if operation.status is None:
    return operation
elif operation.status["state"] in ["PENDING", ...]:
    ...
elif operation.status["state"] in ["CANCELED", ...]:
    ...
```

tamr_client/operation.py
skalish: I wasn't sure if it was best to include all of these.  Looking at `Dataset`, there are 3 required attributes, `url`, 'name`, `key_attribute_names` and 1 optional attribute `description`.  In the docstring, only `url` and `key_attribute_names` are listed under `Args:`.  Is this intentional, and if so is there a pattern that should be followed, or should every attribute be listed?

tamr_client/operation.py
```diff
@@ -0,0 +1,195 @@
...
+    """A Tamr operation
+
+    See https://docs.tamr.com/new/reference/the-operation-object
+
+    Args:
+        url
+    """
+
+    url: URL
+    id: str
```
pcattori, line 36: I don't think we should keep track of `id` separately from `url` since that can duplicate information and potentially become inconsistent

tamr_client/operation.py
skalish: It looks like `apply_options` is used internally in TUC in a few places, mostly `refresh` methods for datasets.  It seems like this is used just to pass along the details of polling frequency and client timeouts.

tamr_client/operation.py
pcattori: We should include all user-facing attributes in the docstring, so I think `Dataset` should be fixed.

skalish: 👍  Once this PR is near merging, `unified.commit` should be updated to return an `Operation`.  I'll fix the docstrings in `Dataset` and `UnifiedDataset` then.

tamr_client/operation.py
```diff
@@ -0,0 +1,188 @@
...
+        url: Operation URL
+        data: Operation JSON data from Tamr server
+    """
+    cp = deepcopy(data)
+    return Operation(
+        url, type=cp["type"], status=cp.get("status"), description=cp.get("description")
+    )
+
+
+def _apply_options(
```
pcattori, line 166: another option we have is to make all functions that return an operation synchronous. Then, if people want asynchronous versions, we can create separate asynchronous functions.

E.g.

```python
def refresh(...) -> Operation:
   ...
   op = _refresh_async(...)
   return tc.operation.wait(op)

def _refresh_async(...) -> Operation:
    ...
```

----

^this design removes the need for a function like `_apply_options` since the synchronous version cannot have options applied at all, and the asynchronous option let's you apply your operation options by calling `tc.operation.wait(...)` yourself:

```python
op = _refresh_async(...)
tc.operation.wait(op, **my_options)
```

ianbakst: I would prefer synchronous operations by default, but perhaps I'm biased by the current operation of the client. Perhaps this would warrant an RFC somewhere down the line?

pcattori: We already did investigation into whether sync or async should be the default in the early TUC days, and the response was overwhelmingly "sync". I don't think this warrants an RFC unless we think something has changed.

tamr_client/dataset/unified.py
```diff
@@ -94,15 +97,26 @@ def _from_json(url: URL, data: JsonDict) -> UnifiedDataset:
     )
 
 
-def commit(session: Session, unified_dataset: UnifiedDataset) -> JsonDict:
-    """Commits the Unified Dataset.
+def apply_changes(session: Session, unified_dataset: UnifiedDataset) -> Operation:
+    """Applies changes to the unified dataset and waits for the operation to resolve
```
pcattori, line 101: nit: wording

```suggestion
    """Applies changes to the unified dataset and waits for the operation to complete
```

PR #398, opened 2020-06-29
Introduce the _types package

pcattori:
# ↪️ Pull Request

Relates to #351 

This PR is meant to be reviewed commit-by-commit. See commit messages for details.

TL;DR:
- Simplify `attribute.type.PrimitiveType` into an `Enum`
- Namespacing clean-up for `attribute` package
- Start moving some types into the `_types` package
- Demote `pandas` to a dev dependency

This PR introduces the pattern of keeping types (i.e. `dataclass`es) in a `_types` package. This greatly simplifies resolution order for (1) type-checking via `mypy` (2) building docs via `sphinx` and (3) running tests via `pytest`. That means that contributors do not have to worry about cyclic imports nor import order to get all of these tools running properly.

Simplified resolution means we can deftly specify that `pandas` is required for type-checking and building docs, but not for users installing TC.

## ✔️ PR Todo


tamr_client/attribute/__init__.py
```diff
@@ -0,0 +1,14 @@
+from . import sub
```
skalish, line 1: It looks like you are using relative imports in the `__init__.py` files.  I think we determined that relative imports are generally okay, but up until now were not being used in the `TC` codebase.  Is there a good reason to use 
`from . import sub` vs `from tamr_client.attribute import sub`?

If that is the case, I think it is worth adding a line to the Internal Imports section of the Style Guide that clarifies when relative imports are appropriate (only in `__init__.py` if I understand this PR).

tamr_client/dataset/dataframe.py
```diff
@@ -22,7 +25,7 @@ class AmbiguousPrimaryKey(Exception):
 def upsert(
     session: Session,
     dataset: Dataset,
-    df: pd.DataFrame,
+    df: "pd.DataFrame",
```
skalish, line 28: This looks like the type/class `pd.DataFrame` is being replaced by a string, since this would require `pandas` to be imported otherwise.  Is this being evaluated in some way when typechecking and building the docs?

tamr_client/attribute/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from . import type
+from ._attribute import (
```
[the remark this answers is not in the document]
pcattori: No technical problem, just a UX one. I wanted users who type `tc.attribute.` and then hit <kbd>TAB</kbd> not to get `tc.attribute.attribute` as a suggestion. I think `_attribute` nicely conveys "if you got here, you've gone too far" unless you're a poweruser.

tamr_client/dataset/dataframe.py
pcattori: Yes, this uses [forward references](https://www.python.org/dev/peps/pep-0484/#forward-references), which both `mypy` and `sphinx_autodoc_typehint` know how to interpret as a type annotation.

tamr_client/attribute/__init__.py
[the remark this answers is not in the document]
skalish: That's fair and I agree that it's worth keeping tab-completion clean. This makes me think it is a pattern that should be repeated for `dataset.py` and any future "eponymous" modules.

tamr_client/attribute/__init__.py
pcattori: We could just delegate to [PEP 8's explanation of imports](https://www.python.org/dev/peps/pep-0008/#imports) i the Style Guide, specifically:

> explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

tamr_client/attribute/__init__.py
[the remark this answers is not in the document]
ianbakst: I guess to add onto this, this will be the start of explicitly hoisting functions and exceptions in the module __init__.py, right?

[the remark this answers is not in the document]
pcattori: We already hoist values from `dataset/*` modules in `dataset/__init__.py`: https://github.com/Datatamer/tamr-client/blob/master/tamr_client/dataset/__init__.py , so this is just a continuation of that.

tamr_client/attribute/__init__.py
skalish: I think this is an example where `from . import sub` and `from tamr_client.attribute import sub` are both reasonable choices.  The former is not too difficult to understand and the latter isn't so verbose as to be problematic.  Personally I prefer the absolute import but I'm fine with either.

pcattori: Changed imports to all be absolute.

PR #420, opened 2020-07-08
Allow creation of an Operation from job resource ID

### LINKED ISSUE #383: Create Operation from Job resource id
# 🙋 feature request

There are lots of jobs that can be kicked off via custom API requests. The response to those APIs usually includes the job id for the operation that has been kicked off, but the response is a 200 code (rather than 204). As a result, the current Operation.from_request() method cannot take that output to create an Operation object.

I would like a way to take a job resource id, to retrieve a job from Tamr and store it as an Operation object so that I can make use of all the nice methods within the Operation object.

## 🤔 Expected Behavior

I'm expecting something like this:
```
res = client.post("../../recipe/modules/{}/job/updateSourceList".format(module_id))
job_id = res.json()["documentId"]["id"]
op = Operation.from_resource_id(client, job_id)
assert op.succeeded()
```

## 😯 Current Behavior
Currently, I would have to submit the job via `client.post()`, and then create my own job poller, which is doable, but annoying.

## 💁 Possible Solution

With the job id, it should be possible to get the same job info that we currently get when we receive a 204 response from other requests. So then it should follow that we could generate an operation in the same way that the `from_response` method works with just a little extra work, possibly hitting the dataset API to retrieve the job info.

## 🔦 Context

I was creating a workflow using the GR module, and realized there was no easy way to kick off a job via non-versioned apis and then use the Operation object gracefully. While versioned APIs for GRs (and modules more broadly) will come out eventually and then added explicitly into this library, I suspect there will always be edge cases where a job needs to be kicked off via internal apis for cutting edge features. So I figured it would be best to have this workaround functional to support those cases.

## 💻 Examples

I have no additional examples to provide other than the one in the expected behavior section above.

### LINKED ISSUE #419: tc.operation.from_resource_id
# 🐛 bug report

Would fix #383 

## 🌍 Your Environment

| Software          | Version(s) |
| ----------------- | ---------- |
| tamr-unify-client |
| Tamr server       |
| Python            |
| Operating System  |

skalish:
# ↪️ Pull Request

This PR adds functionality to both TC and TUC allowing a user to fetch an operation from a Tamr instance by ID. This ID matches the Job ID that appears on the Jobs page in the Tamr UI.
- In TC the user-facing function `from_resource_id` has been added to the `operation` module.  
- In TUC a class method `from_resource_id` has been added to the `Operation` collection.

Closes #419 and closes #383 

## 💻 Examples

TC:
```
import tamr_client as tc

session = tc.Session(...)
job_id = "42"

op = tc.operation.from_resource_id(session, instance, job_id)
```

TUC:
```
from tamr_unify_client import Client, Operation

client = Client(...)
job_id = "42"
op = Operation.from_resource_id(client, job_id)
```
## ✔️ PR Todo

### DIFF
--- a/tamr_client/operation.py
+++ b/tamr_client/operation.py
@@ -77,7 +77,7 @@ def from_resource_id(
         resource_id: The ID of the operation
     """
     url = URL(instance=instance, path=f"operations/{resource_id}")
-    r = session.get(str(url), headers={"Accept": "application/json"})
+    r = session.get(str(url))
     return _from_response(instance, r)
 
 
--- a/tamr_client/operation.py
+++ b/tamr_client/operation.py
@@ -68,6 +68,19 @@ def succeeded(operation: Operation) -> bool:
     return operation.status is not None and operation.status["state"] == "SUCCEEDED"
 
 
+def from_resource_id(
+    session: Session, instance: Instance, resource_id: str
+) -> Operation:
+    """Get operation by ID
+
+    Args:
+        resource_id: The ID of the operation
+    """
+    url = URL(instance=instance, path=f"operations/{resource_id}")
+    r = session.get(str(url), headers={"Accept": "application/json"})
+    return _from_response(instance, r)
+
+
 def _from_response(instance: Instance, response: requests.Response) -> Operation:
     """
     Handle idiosyncrasies in constructing Operations from Tamr responses.
--- a/tamr_unify_client/operation.py
+++ b/tamr_unify_client/operation.py
@@ -30,7 +30,7 @@ class Operation(BaseResource):
         :rtype: :class:`~tamr_unify_client.operation.Operation`
         """
         url = f"operations/{resource_id}"
-        response = client.get(url, headers={"Accept": "application/json"}).successful()
+        response = client.get(url).successful()
         return Operation.from_response(client, response)
 
     @classmethod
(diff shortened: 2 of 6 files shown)


tamr_unify_client/operation.py
```diff
@@ -18,6 +18,21 @@ class Operation(BaseResource):
     def from_json(cls, client, resource_json, api_path=None):
         return super().from_data(client, resource_json, api_path)
 
+    @classmethod
+    def from_resource_id(cls, client, resource_id):
+        """Get an operation by resource ID.
+
+        :param client: Delegate underlying API calls to this client.
```
skalish, line 25: The docstring and typehinting are not in compliance with the current standards.  I ran into difficulties trying to refer to the `Client` and `Operation` classes/types.  I'm not sure if there is a circular import issue or something like that.

pcattori: Type annotations for TUC are best-effort and not required, so this is ok.

tamr_client/operation.py
```diff
@@ -68,6 +68,19 @@ def succeeded(operation: Operation) -> bool:
...
+def from_resource_id(
+    session: Session, instance: Instance, resource_id: str
+) -> Operation:
+    """Get operation by ID
+
+    Args:
+        resource_id: The ID of the operation
+    """
+    url = URL(instance=instance, path=f"operations/{resource_id}")
+    r = session.get(str(url), headers={"Accept": "application/json"})
```
pcattori, line 80: I don't think the `headers` part is necessary given that we don't use that in other `session.get` requests in TC.

skalish: That's right. I'll drop those.

PR #421, opened 2020-07-10
Transformations

### LINKED ISSUE #417: Design for Transformations
# 💬 RFC

Now that we have versioned APIs for transformations, we can add them to tamr-client. This is useful for things like project migration and version control of transformations.

Proposal:
- add data class for the transformations of a project
- add get and replace transformations function to the project module

## 🔦 Context

The APIs GET /v1/projects/{project}/transformations and PUT /v1/projects/{project}/transformations use json in the following format:
```
{
  "parameterized": [
    {
      "datasets": [],
      "transformation": "//all datasets\nSELECT *, 1 as one;"
    },
    {
      "datasets": [
        {
          "name": "people_tiny.csv",
          "datasetId": "unify://unified-data/v1/datasets/18",
          "relativeDatasetId": "datasets/18"
        }
      ],
      "transformation": "//just people\nSELECT *, 1 as one;"
    }
  ],
  "unified": [
    "//unified 1\nSELECT *, 1 as one;",
    "//unifed 2\nSELECT *, 1 as one;"
  ]
}
```

## 💻 Examples

Dataclass Option 1)
```
class Transformations:
    input_scope: input_scope: Optional[List[Tuple[List[Dataset], str]]] = None
    unified_scope: Optional[List[str]] = None
```

Dataclass Option 2)
```
class InputTransformation:
    script: str
    datasets: Optional[List[Dataset]] = None
    
class Transformations:
    input_scope: Optional[List[InputTransformation]] = None
    unified_scope: Optional[List[str]] = None
```    

Functions:

```
def transformations(project: Project) -> Transformations
""" Retrieves all transformations for a project"""
```
```
def replace_transformations(project: Project, transformations: Transformations) -> requests.Response
""" Replaces the current transformations of a project with a new group"""
```

keziah-tamr:
# ↪️ Pull Request

Resolves #417 

Add basic functionality for Transformations:
New functions:
- transformations.get_all(): get transforms from project
- transformations.replace_all(): replace transforms of a project
New dataclasses:
- Transformations
- InputTransformation

~New exceptions: transformations.InvalidInputDataset, transformations.LintingError~

## 💻 Examples

```
import tamr_client as tc
session = tc.session.from_auth('username', 'password')
instance = tc.instance.Instance(host="localhost", port=9100)
project1 = tc.project.from_resource_id(session, instance, id='1')
print(tc.transformations.get_all(session, project1))

```
```
import tamr_client as tc
session = tc.session.from_auth('username', 'password')
instance = tc.instance.Instance(host="localhost", port=9100)
project1 = tc.project.from_resource_id(session, instance, id='1')
dataset3 = tc.dataset.from_resource_id(session, instance, id='3')
new_input_tx = tc.InputTransformation("SELECT *, upper(name) as name;", [dataset3])
all_tx = tc.Transformations(
input_scope=[new_input_tx], unified_scope=["SELECT *, 1 as one;"]
)
tc.transformations.replace_all(session, project1, all_tx)
```

## ✔️ PR Todo


tamr_client/transformations.py
```diff
@@ -0,0 +1,150 @@
+import requests
+from simplejson.errors import JSONDecodeError
```
pcattori, line 2: Can you make this a [json.JSONDecodeError](https://docs.python.org/3/library/json.html#json.JSONDecodeError) using the standard library's `json` module rather than `simplejson`? We are trying to remove the dependency on `simplejson` soon.

You should be able to remove the stub you made for `simplejson` if we switch to `json`.

tamr_client/transformations.py
```diff
@@ -0,0 +1,150 @@
...
+
+from tamr_client import dataset, response
+from tamr_client._types import Instance, JsonDict, Project, Session
+from tamr_client._types.transformations import InputTransformation, Transformations
```
pcattori, line 6: You should access all types from the `_types` package via `from tamr_client._types import <type>`. Looks like you correctly setup the `tamr_client/_types/__init__.py` to do this, so you should just be able to combine the import statements here.

tamr_client/transformations.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class InvalidInputDataset(TamrClientException):
+    """Raised when a Dataset within a InputTransformation
+    is not an input dataset of the target project."""
+
+    pass
+
+
+class LintingError(TamrClientException):
```
pcattori, line 17: Stylistically, our exception names omit words like "exception" or "error" since those are implied by them being exceptions in the first place, but not sure if I can think of a non-awkward wording otherwise. Would something slightly more specific like `LintingFailed` make sense here?

E.g. instead of `AlreadyExistsError` we just say `AlreadyExists`

Not a big deal if we can't come up with something better.

tamr_client/transformations.py
```diff
@@ -0,0 +1,150 @@
...
+    Example:
+        >>> import tamr_client as tc
+        >>> session = tc.session.from_auth('username', 'password')
+        >>> instance = tc.instance.Instance(host="localhost", port=9100)
+        >>> project1 = tc.project.from_resource_id(session, instance, id='1')
+        >>> print(tc.transformations.get_all(session, project1))
+    """
+    r = session.get(
+        f"{project.url}/transformations",
+        headers={"Content-Type": "application/json", "Accept": "application/json"},
```
pcattori, line 100: Based on other usages in the codebase, these headers are unnecessary:

```suggestion
```

tamr_client/transformations.py
```diff
@@ -0,0 +1,150 @@
...
+    body = _to_json(tx)
+    r = session.put(
+        f"{project.url}/transformations",
+        json=body,
+        headers={"Content-Type": "application/json", "Accept": "application/json"},
+    )
+    if r.status_code == 400:
+        try:
+            r_json = r.json()
+            if r_json["class"] == "java.lang.IllegalArgumentException":
```
pcattori, line 140: How sure are we that this error class will always correspond to an invalid input dataset? Same goes for the `javax.ws.rs.BadRequestException`. Could that be caused by something other than linting?

In general, we've interpreted HTTP error codes as custom errors, but that's mostly because those errors are consistent at the protocol-level, whereas this depends on the implementation of our Tamr server (which changes much more rapidly than the HTTP spec 😛 ).

tests/tamr_client/test_transformations.py
```diff
@@ -0,0 +1,139 @@
...
+    responses.add(responses.GET, str(dataset_url), json=dataset_json)
+
+    # test
+    s = utils.session()
+    instance = utils.instance()
+    project = tc.project.from_resource_id(s, instance, "1")
+
+    transforms = tc.transformations.get_all(s, project)
+
+    assert isinstance(transforms, tc.Transformations)
```
pcattori, line 31: Type checking will already guarantee correctness here, no need to include this in the runtime tests.

```suggestion
```

tests/tamr_client/test_transformations.py
```diff
@@ -0,0 +1,139 @@
...
+        tc.transformations.replace_all(s, project, transforms)
+
+    with pytest.raises(tc.transformations.LintingError):
+        tc.transformations.replace_all(s, project, transforms)
+
+    with pytest.raises(HTTPError):
+        tc.transformations.replace_all(s, project, transforms)
+
+    with pytest.raises(HTTPError):
+        tc.transformations.replace_all(s, project, transforms)
```
pcattori, line 139: Is this duplicate block intentional? If so, add a comment to explain the intent.

tamr_client/transformations.py
keziah-tamr: @pcattori  Yeah I was debating this. I'm not that sure that these will be stable, is this something we can ask a DEV about? Two very different and common errors come back as code 400. Without it, we need to leave these as generic expectations.

pcattori: This is a place where precision matters, so could we check the stacktrace for more info to determine if this is indeed the common exception we are expecting?

keziah-tamr: LintingErrors should be very common. Invalid dataset is probably a rare one. The stacktrace on the LintingError still feels a bit generic, not sure what piece to focus on. What do you think of doing a regex match on the message itself?

docs/beta/transformations.rst
```diff
@@ -0,0 +1,13 @@
+Transformations
+===============
+
+.. autofunction:: tamr_client.transformations.get_all
+.. autofunction:: tamr_client.transformations.replace_all
+
+Exceptions
```
pcattori, line 7: Remove these

tamr_client/transformations.py
```diff
@@ -0,0 +1,123 @@
...
+        ... input_scope=[new_input_tx],
+        ... unified_scope=["SELECT *, 1 as one;"]
+        ... )
+        >>> tc.transformations.replace_all(session, project1, all_tx)
+    """
+    body = _to_json(tx)
+    r = session.put(f"{project.url}/transformations", json=body)
+
+    response.successful(r)
+    return r
```
pcattori, line 123: nit: You can combine these:

```suggestion
    return response.successful(r)
```

Issue #423, opened 2020-07-29
TESTING


pcattori:
# 💬 RFC

```python
import pytest
import responses

import tamr_client as tc
from tests.tamr_client import utils

# /fake_json/test_my_test/
# - 1.json # method, path...
# - 2.ndjson
# - 3.json

# 1.json
{
    method:
    path: /
    data: {...}
}
# ...

@fake_tamr
def test_my_test(session):
    jlkj
    jkllkj
    lkjljk
    
    pass
```

skalish:
The naming of these testing resources as `fake` has now come up a few times in conversation.  "Fake" is a generally accessible word, but as far as I can gather it has a pretty specific definition in the context of ["test doubles"](https://blog.pragmatists.com/test-doubles-fakes-mocks-and-stubs-1a7491dfa3da), where fakes are test objects with working (but limited) implementations.  In the utilities introduced in #424, **fake** seems to be an accurate descriptor for the objects returned by `fake.session`, `fake.dataset`, etc. however, the use of the `responses` library is better described as setting up a **stub** server "that holds predefined data and uses it to answer calls during tests".

I don't think there's an obvious answer to how these utilities should be named, but it's probably worth hashing it out and recording a deliberate decision in an ADR.  My main motivation here is being generous towards TC contributors encountering this topic for the first time (I include myself here), since they could benefit if we are thoughtful about the terminology.

pcattori:
My interpretation:

*Mocking* a server means *faking* its responses. I don't think the issue is "mock" vs "fake", but rather "server" vs "response".

If you are faking a server, call it `fake.server`; if you are faking responses, call it `fake.response`; if you are faking JSON data call it `fake.json`.

So I still think the current terminology is appropriate.

pcattori:
For rationale for faking the server in integration tests, see https://kentcdodds.com/blog/stop-mocking-fetch .

PR #424, opened 2020-07-30
Utilities for fake resources

pcattori:
# ↪️ Pull Request

Relates to #423 

## 💻 Examples

See 3rd commit converting `test_project.py` or 6th commit converting `test_transformations.py`

## ✔️ PR Todo
- [x] Add stub for `responses.RequestsMock`
- [x] Add docstrings for `fake` module and `fake.json` decorator
- [x] Allow fake JSON to specify API path (instead of full URL)
- [x] Add docs to contributor guide detailing how tests should be written


tests/tamr_client/test_project.py
```diff
@@ -1,32 +1,24 @@
...
 import tamr_client as tc
-from tests.tamr_client import utils
+from tests.tamr_client import fake, utils
 
 
-@responses.activate
+@fake.json
 def test_from_resource_id_mastering():
     s = utils.session()
     instance = utils.instance()
```
pcattori, line 10: Once we move the utilities into the `fake` module:

```suggestion
    s = fake.session()
    instance = fake.instance()
```

tests/tamr_client/fake.py
```diff
@@ -0,0 +1,33 @@
...
+import responses
+
+tests_tc_dir = (Path(__file__) / "..").resolve()
+fake_json_dir = tests_tc_dir / "fake_json"
+
+
+def _to_kwargs(fake):
+    req = fake["request"]
+    resp = fake["response"]
+    return dict(method=req["method"], url=req["url"], **resp)
```
pcattori, line 15:
```suggestion
    return dict(**req, **resp)
```

skalish:
This looks pretty cool.  I'm still trying to wrap (ha!) my head around details of `wraps`, but this seems like a nice way to make the responses library less imposing.

tests/tamr_client/test_transformations.py
```diff
@@ -40,71 +24,22 @@ def test_get_all():
...
     transforms.input_scope.pop(1)
+    tc.transformations.replace_all(s, project, transforms)
 
-    responses.add(
-        responses.PUT, str(tx_url), json=tc.transformations._to_json(transforms)
-    )
-
-    r = tc.transformations.replace_all(s, project, transforms)
-
-    posted_tx = tc.transformations._from_json(s, project.url.instance, r.json())
```
pcattori, line 73: Author's note: the rest of this test function just tests serialization, which we already test via `test_get_all`. Users should only use the response returns from `tc.tranformations.replace_all` for debugging purposes. If users want the modified transformations, they should just call `tc.transformations.get_all` after.

Alternatively, we could modify `tc.transformations.replace_all` to return the modified.transformations

pcattori:
@clint-richardson @keziah-tamr @along528 : In case you're interested or it's relevant to `tamr-toolbox`, we are exploring simplifications and improvements to our test suite.

pcattori:
> Is there a way to simulate successive calls of the same api endpoint?

@ianbakst : Yes! The fake JSON file is a list of request/response pairs. Simply add another request/response to the file. Multiple requests to the same endpoint are allowed.

One feature we can add later is to specify `ndjson` instead of `json` for the response in the fake JSON file and make sure we handle that correctly in `@fake.json`. But we can do that once we modify a test that requires streaming JSON.

skalish:
I see the very nice documentation in `fake.py`, but I think it would also be worth having a short How-To in the contributor guide to introduce this method of writing tests.  Beyond that, I would love for us to get feedback from the field and then get this merged in.

skalish:
After some reading, I think I understand how the function `fake.json` works and why it is defined in the way it is.  I’m least sure about the use of `wraps`: is it correct that the only effect of the line `@wraps(test_fn)` is to make the returned function `wrapper` look like (i.e. have the same `__name__`, `__module__`, `__docs__` properties as) the function that the `@fake.json`  decorator is applied to?  Or does it affect the basic functionality in some other way?

pcattori:
> is it correct that the only effect of the line `@wraps(test_fn)` is to make the returned function wrapper look like (i.e. have the same `__name__`, `__module__`, `__docs__` properties as) the function that the `@fake.json` decorator is applied to?

That's right. `@wraps` doesn't do anything conceptually. Just does bookkeeping on function metadata like `__name__` to preserve the metadata of the original function in the new one.

PR #428, opened 2020-08-06
TC: Move tc.attribute.from_dataset_all to tc.dataset.attributes

### LINKED ISSUE #418: Move tc.attribute.from_dataset_all to tc.dataset.attributes

skalish:
# ↪️ Pull Request
This PR moves the function `tamr_client.attribute.from_dataset_all` to `tamr_client.dataset.attributes`.  Organizing the code to group together functionality around what can be asked about or what can be done with a dataset is more natural than grouping together the different ways a user might get attributes (e.g. `from_dataset`, `from_project`).

Closes #418 .

This PR also makes a minor modification to the dev task `format` so that the `black` flag `--diff` can be used.

## 💻 Examples

```python
s = fake.session()
dataset = fake.dataset()

attrs = tc.dataset.attributes(s, dataset)

row_num = attrs[0]
attr_1 = attrs[1]
attr_2 = attrs[2]
```

## ✔️ PR Todo


skalish:
I could roll updating the `dataset` tests to use the `fake` utility into this PR if we want to continue that conversion. (#423)

pcattori:
> I could roll updating the `dataset` tests to use the `fake` utility into this PR if we want to continue that conversion. (#423)

@skalish : Yes, we should be migrating tests to use `fake` utility as we touch them. This will require adding `ndjson` capabilities to `fake.json`. I'm envisioning that the fake JSON file can specify `{..., "response": { "ndjson": [...] } }` and the list supplied in `fake_json["response"]["ndjson"]` gets used to set up `responses` with a response stream.

tests/tamr_client/dataset/test_dataset.py
```diff
@@ -30,3 +32,23 @@ def test_from_resource_id_dataset_not_found():
...
+
+
+@responses.activate
+def test_attributes():
+    s = fake.session()
+    dataset = fake.dataset()
+
+    attrs_url = replace(dataset.url, path=dataset.url.path + "/attributes")
+    attrs_json = utils.load_json("attributes.json")
+    responses.add(responses.GET, str(attrs_url), json=attrs_json, status=204)
```
pcattori, line 44: Let's include adding `ndjson` compatibility to `fake.json` so we can rewrite this without explicitly relying on `responses`.

skalish: I think the `GET /dataset/{dataset_id}/attributes` endpoint returns a JSON list and not a stream.  I have taken a crack at adding `ndjson` compatibility to `fake.json`, but it doesn't seem to be applicable to the `attributes` functionality.  I could modify the testing for `record.stream` which would use this, but maybe that should be a separate PR.

pcattori: Ah good point. Yes, we should include `ndjson` capability in a PR that touches streaming. Let's leave out of this one.

pcattori: Note: we should still convert this test to use `fake.json`

skalish:
Still working on migrating the `attribute` tests to use `fake.json`, so a merge should wait a bit longer.

skalish:
I’m getting stuck on one last thing for updating the attribute testing.  With the changes I’ve made, there is still a heavy reliance on loading `tests/tamr_client/data/attribute.json` and `tests/tamr_client/data/attributes.json`.  In one place, this is used in the `test_from_json` function that is not about interacting with the server and I’m not sure how to change that to not use this file.  On the other hand, it is also used in the tests to compare the response from the server to the `Attribute` that is generated using `_from_json` with the file’s contents https://github.com/Datatamer/tamr-client/blob/0a1b54d5e0f158f22cc0ccbea5d7b9914548524c/tests/tamr_client/attribute/test_attribute.py#L61  I think the way the project tests are written is neater, where some attributes are checked against hard-coded values https://github.com/Datatamer/tamr-client/blob/0a1b54d5e0f158f22cc0ccbea5d7b9914548524c/tests/tamr_client/test_project.py#L13-L15 but without resolving `test_from_json` it wouldn’t let us drop those (now kind of redundant) JSON files.

skalish:
The tests that make use of `fake.json` no longer also load a JSON from `tests/tamr_client/data`, requiring redundant data in multiple locations.  The two tests that do not use `fake.json` still load `attributes.json`, so that file has not been removed (it is used by tests in `tests/tamr_client/attribute/test_type.py`).

tests/tamr_client/fake.py
```diff
@@ -101,3 +101,13 @@ def transforms() -> tc.Transformations:
         ],
         unified_scope=["//Comment\nSELECT *;"],
     )
+
+
+def attribute() -> tc.Attribute:
+    return tc.Attribute(
+        url=tc.URL(path="datasets/1/attributes/RowNum"),
+        name="RowNum",
+        type=tc.attribute.type.Array(tc.attribute.type.STRING),
```
pcattori, line 110: nit: I think `Array<String>` is the default type for attributes, so you can omit.

Even if you don't omit, we should have a `DEFAULT` alias for it somewhere that you can use. Though maybe `DEFAULT` is not a good enough name 🤷

skalish: It looks like `type` is not an optional argument to `tc.Attribute` if that is what you mean by omitting.  I can pass `type=tc.attribute.type.DEFAULT` and that works.

Do we want `type` to be optional in instantiating a `tc.Attribute` and default to `DEFAULT`?

pcattori: Not sure. Let's just use `DEFAULT` explicitly here. We can decide on making `type` optional in follow-on work.

PR #431, opened 2020-08-07
update docs for uploading pandas dataframe as Tamr dataset

### LINKED ISSUE #426: Doc suggestion of using `astype(str)` on DataFrames before upload to Tamr is bad for `NaN`s
The `pandas` method `astype(str)` casts all values in a DataFrame to strings, which allows them to be successfully uploaded to Tamr.  However, special values like `NaN` and the Python `None` will be converted into strings (e.g. "NaN"), introducing non-standard nulls to your Tamr dataset.

An alternative is casting with `astype(object)`, which will preserve these special values.  I'm unsure if there is a downside to this, but I think it is probably a better practice overall.

Related to #323, maybe #373

olivito:
# ↪️ Pull Request

Resolves #426:

The pandas method `astype(str)` casts all values in a DataFrame to strings, which allows them to be successfully uploaded to Tamr. However, special values like `NaN` and the Python `None` will be converted into strings (e.g. `"NaN"`), introducing non-standard nulls to your Tamr dataset.

Using `dtype=object` when creating a dataframe with `pd.read_csv()` correctly preserves null values.

I'm not sure what the best solution is for an already-existing dataframe, but the docs should be updated to reflect the better practice.

### DIFF
--- a/docs/user-guide/pandas.md
+++ b/docs/user-guide/pandas.md
@@ -97,19 +97,26 @@ stored in the removed attributes.
 ## Upload Dataframe as Dataset
 
 ### Create New Dataset
-To create a new dataset and upload data, the convenience function `dataset.create_from_dataframe()` can be used. 
+To create a new dataset and upload data, the convenience function `datasets.create_from_dataframe()` can be used. 
 Note that Tamr will throw an error if columns aren't generally formatted as strings. (The exception being geospatial
 columns. For that, see the geospatial examples.)
 
-In order to achieve this, the following code will transform the column types to string.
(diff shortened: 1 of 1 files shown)


pcattori:
@olivito Pro-Tip: If you use "Resolves" or "Fixes" before the referenced issue, the issue will be auto-closed when this gets merged. E.g. `Resolves #32`

olivito:
> @olivito Pro-Tip: If you use "Resolves" or "Fixes" before the referenced issue, the issue will be auto-closed when this gets merged. E.g. `Resolves #32`

Good to know.  I've changed the PR description.  Hopefully it gets picked up

olivito:
I've squashed the commits and provided a more informative message

PR #432, opened 2020-08-12
TC: Add ndjson and request payload checking support to fake.json + update tests for tc.record

skalish:
# ↪️ Pull Request

Relates to #423 

## ✔️ PR Todo

### DIFF
--- a/tests/tamr_client/fake.py
+++ b/tests/tamr_client/fake.py
@@ -18,9 +18,9 @@ tests_tc_dir = (Path(__file__) / "..").resolve()
 fake_json_dir = tests_tc_dir / "fake_json"
 
 
-def check_payload(request, correct_payload, status, response_json):
-    if correct_payload is not None:
-        if [x.decode("utf-8") for x in request.body] != correct_payload:
+def check_request_body(request, request_body, status, response_json):
+    if request_body is not None:
+        if [x.decode("utf-8") for x in request.body] != request_body:
             raise Exception("placeholder exception")
     return status, {}, response_json
 
@@ -34,22 +34,29 @@ def add_response(rsps, fake):
         path = req.get("path")
         url = "http://localhost/api/versioned/v1/" + path
 
-    ndjson = resp.pop("ndjson", None)
+    # Get response body from either ndjson or json
+    ndjson = resp.get("ndjson")
     if ndjson is not None:
         resp["body"] = "\n".join((dumps(line) for line in ndjson))
+    else:
+        resp["body"] = dumps(resp.get("json"))
 
-    payload = req.pop("payload", None)
-    callback = partial(
-        check_payload,
-        correct_payload=[dumps(x) for x in payload] if payload is not None else None,
-        status=resp["status"],
-        response_json=resp.get("body") or dumps(resp.get("json")),
-    )
+    # Get expected request body from ndjson
+    payload = req.get("ndjson")
+    if payload is not None:
+        req["body"] = [dumps(x) for x in payload]
+    else:
+        req["body"] = None
 
     rsps.add_callback(
         method=req["method"],
         url=url,
-        callback=callback,
+        callback=partial(
+            check_request_body,
+            request_body=req["body"],
+            status=resp["status"],
+            response_json=resp["body"],
+        ),
     )
 
 
--- a/tests/tamr_client/fake.py
+++ b/tests/tamr_client/fake.py
@@ -8,10 +8,12 @@ from functools import partial, wraps
 from inspect import getfile
 from json import dumps, load, loads
 from pathlib import Path
+from typing import Dict, Tuple
 
 import responses
 
 import tamr_client as tc
+from tamr_client._types import JsonDict
 
 
 tests_tc_dir = (Path(__file__) / "..").resolve()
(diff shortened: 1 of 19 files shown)


tests/tamr_client/fake.py
```diff
@@ -27,7 +27,16 @@ def _to_kwargs(fake):
...
+    if "ndjson" in resp.keys():
+        ndjson = resp.pop("ndjson")
+        return dict(
+            method=req["method"],
+            url=url,
+            body="\n".join((dumps(line) for line in ndjson)),
+            **resp
+        )
+    else:
+        return dict(method=req["method"], url=url, **resp)
```
pcattori, line 39: You can simplify this logic by using the second argument of `pop`:

```suggestion
    ndjson = resp.pop("ndjson", None)
    if ndjson is not None:
        resp["body"] = "\n".join((dumps(line) for line in ndjson))

    return dict(method=req["method"], url=url, **resp)
```

tests/tamr_client/fake_json/dataset/test_record/test_stream.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "request": {
+            "method": "GET",
+            "path": "datasets/1/records"
+        },
+        "response": {
+            "ndjson": [
+                {"primary_key": 1}, {"primary_key": 2}
```
pcattori, line 9: nit: can we newline-separate these items to be closer to the `ndjson` format?

```suggestion
                {"primary_key": 1}, 
                {"primary_key": 2}
```

tests/tamr_client/fake_json/categorization/test_project/test_manual_labels.json
```diff
@@ -5,6 +5,7 @@
       "path": "projects/2/unifiedDataset"
     },
     "response": {
+      "status": 200,
```
skalish, line 8: For all of the `test_x.json` files, I filled in the status according to what is described on the swagger docs for Tamr, however, I am not confident these are correct in actual operation.  For instance, I think many that are listed as `200` are actually `204`.  This might not matter, as long as these tests are not relying on the response codes.

tests/tamr_client/fake.py
```diff
@@ -18,7 +18,27 @@
...
+
+
+def check_request_body(request, request_body, status, response_json):
+    if isinstance(request_body, list):
+        caught_body = [loads(x.decode("utf-8")) for x in request.body]
+        if caught_body != request_body:
+            raise WrongRequestBody(caught_body)
+    elif request_body is not None:
+        caught_body = loads(request.body.decode("utf-8"))
+        if caught_body != request_body:
```
skalish, line 36: There is an intentional decision here to compare these as JSON objects instead of strings, i.e. to do `loads(request.body.decode("utf-8")` rather than `dumps(request_body)`.  In two strings dumped from two effectively identical JSON dictionaries, a misleading inequality can result from the arbitrary ordering of elements contained in the dictionaries.

tests/tamr_client/fake.py
```diff
@@ -27,7 +47,28 @@ def _to_kwargs(fake):
...
+    elif req.get("json") is not None:
+        req["body"] = req["json"]
+
+    rsps.add_callback(
+        method=req["method"],
+        url=url,
+        callback=partial(
+            check_request_body,
+            request_body=req.get("body"),
+            status=resp["status"],
```
pcattori, line 68: _Does_ every response need a status? The `responses` API suggests "no". What's the rationale for asserting this?

skalish: The `callback` function parameter of `responses.add_callback` must return a tuple of [`(status, headers, body)`](https://github.com/getsentry/responses/blob/c21ba81b0053a887d412227acadb8d65b331db94/responses.py#L389-L395).  The implementation of `request.Response.raise_for_status()` requires this to be an `int`.  One way of handling this would be to pass `resp.get("status", 200)` (or some other default value), but this seems slightly problematic since it is causing the mock response to return a status that might not be reflective of the actual server response status.

I was under the impression that any real server response should return a status code, but I'm not basing that on much besides the above.  If we choose to use a default value, maybe `0` is reasonable, as that is what `urllib3` does [here](https://github.com/urllib3/urllib3/blob/f0c43419e45db17a3c5f287be80aa0d14e2f58c6/src/urllib3/response.py#L355).

pcattori: Ah ok. Weird that the status is optional for `responses.add`, but required for `responses.add_callback`. Makes sense.

tests/tamr_client/fake.py
```diff
@@ -27,7 +47,28 @@ def _to_kwargs(fake):
...
-    return dict(method=req["method"], url=url, **resp)
+    # Get response body from either ndjson or json
+    if resp.get("ndjson") is not None:
+        resp["body"] = "\n".join((dumps(line) for line in resp["ndjson"]))
+    elif resp.get("json") is not None:
+        resp["body"] = dumps(resp["json"])
+
+    # Get expected request body from ndjson
+    if req.get("ndjson") is not None:
+        req["body"] = [x for x in req["ndjson"]]
```
pcattori, line 58: Unnecessary list comprehension:

```suggestion
        req["body"] = req["ndjson"]
```

tests/tamr_client/fake.py
```diff
@@ -27,7 +47,28 @@ def _to_kwargs(fake):
...
+    elif req.get("json") is not None:
+        req["body"] = req["json"]
+
+    rsps.add_callback(
+        method=req["method"],
+        url=url,
+        callback=partial(
```
pcattori, line 65: One organizational alternative is to have `check_request_body` just take the `request` and `expected_body` as args and raise exception if they don't match. Then the rest of the logic (passing through `status` and `response_json`) can be encapsulated in another function, say `_callback`:

```python
def _callback(request, expected_body, status, response_json):
    check_request_body(request, expected_body)
    return status, {}, response_json
```

I bring this up only because the name `check_request_body` implies that this function should *only* perform a check and have no side-effects or need to take the `status` or `response_json` as args.

Another nit: we should probably put a leading prefix on `_check_response_body` in any case.

tests/tamr_client/fake.py
```diff
@@ -27,7 +51,28 @@ def _to_kwargs(fake):
...
+        req["body"] = req["ndjson"]
+    elif req.get("json") is not None:
+        req["body"] = req["json"]
+
+    rsps.add_callback(
+        method=req["method"],
+        url=url,
+        callback=partial(
+            _callback,
+            request_body=req.get("body"),
```
pcattori, line 71:
```suggestion
            expected_body=req.get("body"),
```

skalish:
Note: One more commit incoming with docstrings addressing the remaining comments.

PR #434, opened 2020-08-13
Add tc.instance.version() function to return version of Tamr

### LINKED ISSUE #427: Method to fetch the Tamr software version from "/api/versioned/service/version".
# 🙋 feature request

## 🤔 Expected Behavior

```
tamr_software_version = tamr.service_version.version

print(f"tamr_software_version {tamr_software_version}")


```

## 🔦 Context

We often perform Tamr software version checks, typically
- to warn users of potential version mismatch issues
- implement version specific features

## 💻 Examples
```
if version_geq(tamr.service_version.version, "2020.004.0"):
   # fetch projects
else: 
   # fetch projects and modules
```

ianbakst:
# ↪️ Pull Request

There is a function now: `tc.instance.version()` which returns the Tamr version of the instance.
Resolves #427 

## 💻 Examples
```python
s = tc.session.from_auth(tc.UsernamePasswordAuth(username,password))
inst = tc.Instance(host,port)
version = tc.instance.version(s,inst)
```
## ✔️ PR Todo


tamr_client/instance.py
```diff
@@ -10,3 +10,20 @@ def origin(instance: Instance) -> str:
...
+        session: Tamr Session
+        instance: Tamr instance
+
+    Returns: Version
+
+    """
+    return _version(session, instance)["version"]
+
+
+def _version(session: Session, instance: Instance) -> dict:
```
skalish, line 28: I'm not sure this needs to be separated into these two functions.  I think I would prefer it to be one.

ianbakst: I mean, it doesn't NEED to be, but there's more in the response than just the version number itself. Perhaps that would be useful to someone, perhaps it wouldn't, but that's why it's not exposed. I feel like the only argument against not building it this way is that it requires building 1 more function, but it's already built. I would prefer it to be two.

ianbakst: I changed my mind. None of that extra stuff should be accessible in any way.

tamr_client/instance.py
```diff
@@ -10,3 +10,16 @@ def origin(instance: Instance) -> str:
...
+    """Return the Tamr version for an instance.
+
+    Args:
+        session: Tamr Session
+        instance: Tamr instance
+
+    Returns: Version
+
+    """
+    return session.get(f"{origin(instance)}/api/versioned/service/version").json()["version"]
```
pcattori, line 25: Might be nice to add a comment for future contributors:

```suggestion
    # Version endpoints are not themselves versioned by design, but they are stable so they are ok to use here.
    return session.get(f"{origin(instance)}/api/versioned/service/version").json()["version"]
```

PR #134, opened 2019-06-11
Ability to create project

mollysacks:
# ↪️ Pull Request

Resolves #86 

## ✔️ PR Todo

- [x] Added/updated testing for this change
- [x] Included links to related issues/PRs
- [x] Update relevant [docs](https://github.com/Datatamer/unify-client-python/tree/master/docs) + docstrings
- [ ] Update the [CHANGELOG](https://github.com/Datatamer/unify-client-python/blob/master/CHANGELOG.md) under the current `-dev` version:
  - Add changelog entries under any that apply: **BREAKING CHANGES**, **NEW FEATURES**, **BUG FIXES**.
  - Changelog entry format: `[#<issue number>](<link to issue>) <change description>`

PR #435, opened 2020-08-14
Create projects in Tamr

abafzal:
<!---

Create projects (Mastering, Categorization, Schema Mapping) in Tamr. 

# ↪️ Pull Request

Addresses Issue #134 

## 💻 Examples

tc.categorization.project.create(...)
tc.mastering.project.create(...)
tc.schema_mapping.project.create(...)

## ✔️ PR Todo


abafzal:
Can you check if the basic organization/design here makes sense? Also, typechecks are failing but I'm not sure what the errors mean.

tamr_client/project.py
```diff
@@ -12,6 +13,12 @@ class NotFound(TamrClientException):
     pass
 
 
+class Conflict(TamrClientException):
```
pcattori, line 16: We have an existing convention of calling this sort of error and `AlreadyExists` error:

```suggestion
class AlreadyExists(TamrClientException):
```

tamr_client/project.py
```diff
@@ -55,5 +62,56 @@ def _from_json(url: URL, data: JsonDict) -> Project:
         return mastering_project._from_json(url, data)
     elif proj_type == "CATEGORIZATION":
         return categorization_project._from_json(url, data)
+    elif proj_type == "SCHEMA_MAPPING_RECOMMENDATIONS":
+        return schema_mapping_project._from_json(url, data)
     else:
         raise ValueError(f"Unrecognized project type '{proj_type}' in {repr(data)}")
+
+
+def create(
```
pcattori, line 71: This function should be internal-only to encourage users to use specific variants of `create` for their specific project type:

```suggestion
def _create(
```

tamr_client/project.py
```diff
@@ -55,5 +62,56 @@ def _from_json(url: URL, data: JsonDict) -> Project:
...
+        unified_dataset_name = name + "_unified_dataset"
+    data = [
+        {
+            "name": name,
+            "type": proj_type,
+            "unifiedDatasetName": unified_dataset_name,
+            "description": description,
+            "externalId": external_id,
+        }
+    ]
```
pcattori, line 105: I think the [type error](https://github.com/Datatamer/tamr-client/pull/435/checks?check_run_id=985999724) you are seeing is because `data` is a `List`, but later on you try to do `data["relativeId"]`, which is trying to access an element of a list via string key. Did you mean to exclude the `[]` when defining `data`?

```suggestion
    data = {
        "name": name,
        "type": proj_type,
        "unifiedDatasetName": unified_dataset_name,
        "description": description,
        "externalId": external_id,
    }
```

tamr_client/categorization/project.py
```diff
@@ -21,6 +23,39 @@ def _from_json(url: URL, data: JsonDict) -> CategorizationProject:
     )
 
 
+def create(
+    session: Session,
+    instance: Instance,
+    name: str,
+    description: str = None,
+    external_id: str = None,
+    unified_dataset_name: str = None,
```
pcattori, line 32: Optional values should be typed with `Optional`:

```suggestion
    description: Optional[str] = None,
    external_id: Optional[str] = None,
    unified_dataset_name: Optional[str] = None,
```

If you do this^, you should omit the `(Optional)` text in the docstring description since our tooling will automatically mark a field as optional in the docs if its typed as an `Optional`.

tamr_client/project.py
abafzal: changed

tamr_client/project.py
abafzal: changed

tamr_client/project.py
abafzal: updated

tamr_client/categorization/project.py
abafzal: done

tamr_client/categorization/project.py
```diff
@@ -21,6 +25,39 @@ def _from_json(url: URL, data: JsonDict) -> CategorizationProject:
...
+        instance: Tamr instance
+        name: Project name
+        description: Project description
+        external_id: External ID of the project
+        unified_dataset_name: Unified dataset name. If None, will be set to project name + _'unified_dataset'
+
+    Returns:
+        A Categorization project object
+
+    Raises:
```
pcattori, line 47: Missing exception type (I think there are 2 other occurrences of this in the code)

abafzal: removed the exception type from the docstring of this function since the actual exception is raised by the _create() method

tamr_client/categorization/project.py
```diff
@@ -21,6 +25,33 @@ def _from_json(url: URL, data: JsonDict) -> CategorizationProject:
...
+def create(
+    session: Session,
+    instance: Instance,
+    name: str,
+    description: Optional[str] = None,
+    external_id: Optional[str] = None,
+    unified_dataset_name: Optional[str] = None,
+) -> Project:
+    """Create a Categorization project in Tamr
+    Args:
```
skalish, line 37:
```suggestion
    """Create a Categorization project in Tamr
    
    Args:
```

The empty line is needed here for the resulting docs to be formatted properly.  This needs to be changed in the other project creation function docstrings, and in that of `manual_labels` below.

tamr_client/categorization/project.py
pcattori: The issue is that if users look at the generated reference docs, the exception will be missing, so we need to include for all user-facing functions.

One thought: the docstrings seem repetitive, so we could write the docstring as a plain string once and use it for multiple functions via the `__doc__` property. See https://softwareengineering.stackexchange.com/a/327664/94802 . Not sure if I like this approach or not yet, but its an option.

abafzal: I think i prefer being rather verbose than referring the user to docstrings of other functions to figure out what errors could be thrown etc.

tamr_client/project.py
```diff
@@ -55,5 +64,54 @@ def _from_json(url: URL, data: JsonDict) -> Project:
...
+        external_id: External ID of the project
+        unified_dataset_name: Name of the unified dataset
+
+    Returns:
+        Project created in Tamr
+
+    Raises:
+        AlreadyExists: If a project with these specifications already exists.
+    """
+    if not unified_dataset_name:
```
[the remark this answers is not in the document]
abafzal: changed

tamr_client/categorization/project.py
abafzal: updated

abafzal: Will create a separate issue for the testing portion

tamr_client/mastering/project.py
```diff
@@ -9,3 +19,37 @@ def _from_json(url: URL, data: JsonDict) -> MasteringProject:
...
+def create(
+    session: Session,
+    instance: Instance,
+    name: str,
+    description: Optional[str] = None,
+    external_id: Optional[str] = None,
+    unified_dataset_name: Optional[str] = None,
+) -> Project:
+    """Create a Mastering project in Tamr.
+    Args:
```
pcattori, line 33: Needs whitespace for doc formatting:

```suggestion

    Args:
```

tamr_client/project.py
```diff
@@ -55,5 +64,55 @@ def _from_json(url: URL, data: JsonDict) -> Project:
...
+    session: Session,
+    instance: Instance,
+    name: str,
+    project_type: str,
+    description: Optional[str] = None,
+    external_id: Optional[str] = None,
+    unified_dataset_name: Optional[str] = None,
+) -> Project:
+    """Create a project in Tamr.
+    Args:
```
pcattori, line 83: Needs whitespace for doc formatting:

```suggestion

    Args:
```

tamr_client/schema_mapping/project.py
```diff
@@ -0,0 +1,57 @@
...
+def create(
+    session: Session,
+    instance: Instance,
+    name: str,
+    description: Optional[str] = None,
+    external_id: Optional[str] = None,
+    unified_dataset_name: Optional[str] = None,
+) -> Project:
+    """Create a Schema Mapping project in Tamr.
+    Args:
```
pcattori, line 35: Needs whitespace for doc formatting:

```suggestion

    Args:
```

abafzal: updated.

tamr_client/project.py
abafzal: updated

tamr_client/mastering/project.py
abafzal: updated

tamr_client/project.py
abafzal: this doesnt show up in the docs since its an internal function

PR #438, opened 2020-08-17
Add functionality to get, initiate and cancel Tamr backups and restores

### LINKED ISSUE #313: Support for Backup of Tamr
# 🙋 feature request

To run a backup of Tamr and assert its success.

## 🤔 Expected Behavior
```
operation = tamr.backup.refresh()
assert operation.succeeded()
```

## 🔦 Context
We often schedule backups of Tamr deployments.

## 💻 Examples
Tamr's API v1 has the following endpoints to support backup creation.

- `GET /v1/backups` Get all backups
- `POST /v1/backups` Initiate an asynchronous backup operation
- `GET /v1/backups/{backupId}` Get a backup by ID
- `POST /v1/backups/{backupId}:cancel` Cancel a running backup, given its ID

abafzal:
# ↪️ Pull Request

Resolves #313
Add functionality to initiate, get, and cancel Tamr backups and restores. 

## 💻 Examples

```
tamr_client.instance.backup.get(...)
tamr_client.instance.backup.get_by_id(..., backup_id="")
tamr_client.instance.backup.initiate(...)
tamr_client.instance.backup.cancel(..., backup_id="")
```

## ✔️ PR Todo

### DIFF
--- a/tamr_client/_types/backup.py
+++ b/tamr_client/_types/backup.py
@@ -1,0 +1,24 @@
+from dataclasses import dataclass
+
+from tamr_client._types.url import URL
+
+
+@dataclass(frozen=True)
+class Backup:
+    """A Tamr backup
+
+    See https://docs.tamr.com/new/docs/configuration-backup-and-restore
+
+    Args:
+        url
+        resource_id
+        path
+        state
+        error_message
+    """
+
+    url: URL
+    resource_id: str
+    path: str
+    state: str
+    error_message: str
--- a/tamr_client/_types/backup.py
+++ b/tamr_client/_types/backup.py
@@ -11,14 +11,12 @@ class Backup:
 
     Args:
         url
-        resource_id
         path
         state
         error_message
     """
 
     url: URL
-    resource_id: str
     path: str
     state: str
     error_message: str
--- a/tamr_client/_types/restore.py
+++ b/tamr_client/_types/restore.py
@@ -1,0 +1,24 @@
+from dataclasses import dataclass
+
+from tamr_client._types.url import URL
+
+
+@dataclass(frozen=True)
+class Restore:
+    """A Tamr restore
+
+    See https://docs.tamr.com/new/docs/configuration-backup-and-restore
+
+    Args:
+        url
+        resource_id
+        backup_path
+        state
+        error_message
+    """
+
+    url: URL
+    resource_id: str
+    backup_path: str
+    state: str
+    error_message: str
--- a/tamr_client/_types/restore.py
+++ b/tamr_client/_types/restore.py
@@ -11,14 +11,12 @@ class Restore:
 
     Args:
         url
-        resource_id
         backup_path
         state
         error_message
     """
 
     url: URL
-    resource_id: str
     backup_path: str
     state: str
     error_message: str
--- a/tamr_client/backup.py
+++ b/tamr_client/backup.py
@@ -125,3 +125,26 @@ def cancel(session: Session, backup: Backup) -> Backup:
     if r.status_code == 400:
         raise InvalidOperation(cancel_url, r.json()["message"])
     return _from_json(backup.url, response.successful(r).json())
+
+
+def poll(session: Session, backup: Backup) -> Backup:
+    """Poll this backup for server-side updates.
+
+    Does not update the :class:`~tamr_client.backup.Backup` object.
+    Instead, returns a new :class:`~tamr_client.backup.Backup`.
+
+    Args:
+        session: Tamr session
+        backup: Tamr backup to be polled
+
+    Returns:
+        A Tamr backup
+
+    Raises:
+        backup.NotFound: If no backup found at the specified URL
+    """
+    url = backup.url
+    r = session.get(str(url))
+    if r.status_code == 404:
+        raise NotFound(str(url))
+    return _from_json(url, response.successful(r).json())
--- a/tamr_client/backup.py
+++ b/tamr_client/backup.py
@@ -30,7 +30,6 @@ def _from_json(url: URL, data: JsonDict) -> Backup:
     cp = deepcopy(data)
     return Backup(
         url=url,
-        resource_id=cp["relativeId"],
         path=cp["backupPath"],
         state=cp["state"],
         error_message=cp["errorMessage"],
@@ -119,10 +118,10 @@ def cancel(session: Session, backup: Backup) -> Backup:
         backup.NotFound: If no backup found at the specified URL
         backup.InvalidOperation: If attempting an invalid operation
     """
-    r = session.post(f"{backup.url}:cancel")
+    cancel_url = f"{backup.url}:cancel"
+    r = session.post(cancel_url)
     if r.status_code == 404:
-        raise NotFound(f"{backup.url}:cancel")
+        raise NotFound(cancel_url)
     if r.status_code == 400:
-        raise InvalidOperation(f"{backup.url}:cancel", r.json()["message"])
-
+        raise InvalidOperation(cancel_url, r.json()["message"])
     return _from_json(backup.url, response.successful(r).json())
--- a/tamr_client/backup.py
+++ b/tamr_client/backup.py
@@ -1,0 +1,128 @@
+from copy import deepcopy
+from typing import List
+
+from tamr_client import Backup, response
+from tamr_client._types import Instance, JsonDict, Session, URL
+from tamr_client.exception import TamrClientException
+
+
+class InvalidOperation(TamrClientException):
+    """Raised when attempting an invalid operation.
+    """
+
+    pass
+
+
+class NotFound(TamrClientException):
+    """Raised when referencing a backup that does not exist on the server.
+    """
+
+    pass
+
+
+def _from_json(url: URL, data: JsonDict) -> Backup:
+    """Make backup from JSON data (deserialize).
+
+    Args:
+        url: Backup URL
+        data: Backup JSON data from Tamr server
+    """
+    cp = deepcopy(data)
+    return Backup(
+        url=url,
+        resource_id=cp["relativeId"],
+        path=cp["backupPath"],
+        state=cp["state"],
+        error_message=cp["errorMessage"],
+    )
+
+
+def get_all(session: Session, instance: Instance) -> List[Backup]:
+    """Get all backups that have been initiated for a Tamr instance.
+
+    Args:
+        session: Tamr session
+        instance: Tamr instance
+
+    Returns:
+        A list of Tamr backups
+
+    Raises:
+        backup.NotFound: If no backup found at the specified URL
+    """
+    url = URL(instance=instance, path="backups")
+    r = session.get(str(url))
+    if r.status_code == 404:
+        raise NotFound(str(url))
+    backups = [
+        _from_json(URL(instance=instance, path=f'backups/{data["relativeId"]}'), data)
+        for data in response.successful(r).json()
+    ]
+    return backups
+
+
+def by_resource_id(session: Session, instance: Instance, resource_id: str) -> Backup:
+    """Get information on a specific Tamr backup.
+
+    Args:
+        session: Tamr session
+        instance: Tamr instance
+        resource_id: Resource ID of the backup
+
+    Returns:
+        A Tamr backup
+
+    Raises:
+        backup.NotFound: If no backup found at the specified URL
+    """
+    url = URL(instance=instance, path=f"backups/{resource_id}")
+    r = session.get(str(url))
+    if r.status_code == 404:
+        raise NotFound(str(url))
+    return _from_json(url, response.successful(r).json())
+
+
+def initiate(session: Session, instance: Instance) -> Backup:
+    """Initiate a Tamr backup.
+
+    Args:
+        session: Tamr session
+        instance: Tamr instance
+
+    Returns:
+        Initiated backup
+
+    Raises:
+        backup.InvalidOperation: If attempting an invalid operation
+    """
+    url = URL(instance=instance, path="backups")
+    r = session.post(str(url))
+    if r.status_code == 400:
+        raise InvalidOperation(str(url), r.json()["message"])
+    data = response.successful(r).json()
+    return _from_json(
+        URL(instance=instance, path=f'backups/{data["relativeId"]}'), data
+    )
+
+
+def cancel(session: Session, backup: Backup) -> Backup:
+    """Cancel a Tamr backup.
+
+    Args:
+        session: Tamr session
+        backup: A Tamr backup
+
+    Returns:
+        Canceled backup
+
+    Raises:
+        backup.NotFound: If no backup found at the specified URL
+        backup.InvalidOperation: If attempting an invalid operation
+    """
+    r = session.post(f"{backup.url}:cancel")
+    if r.status_code == 404:
+        raise NotFound(f"{backup.url}:cancel")
+    if r.status_code == 400:
+        raise InvalidOperation(f"{backup.url}:cancel", r.json()["message"])
+
+    return _from_json(backup.url, response.successful(r).json())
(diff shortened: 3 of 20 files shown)


tamr_client/instance/backup.py
```diff
@@ -0,0 +1,94 @@
+from tamr_client import response
+from tamr_client._types import Instance, Session, URL
+from tamr_client.exception import TamrClientException
+
+
+class AlreadyCanceled(TamrClientException):
```
abafzal, line 6: @pcattori @skalish would appreciate thoughts on the name of this exception

abafzal:
Added backup functionality. Restore to be added.

tamr_client/instance/backup.py
skalish: What is the result of trying to cancel a cancelled backup in the API?  How does it differ from trying to cancel a failed, finished, or nonexistent backup?  Depending on the answers to this I might prefer something like `NotRunning`.

tamr_client/instance/backup.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class NotFound(TamrClientException):
+    """Raised when referencing a backup that does not exist on the server.
+    """
+
+    pass
+
+
+def get(session: Session, instance: Instance) -> list:
```
skalish, line 20: I think this name should indicate that it will return all backups.  Maybe `get_all`.  It might be worth comparing to how we get a dataset's attributes, which we intentional chose to define as `tc.dataset.attributes`.  The equivalent here would be `tc.instance.backups`.

tamr_client/instance/backup.py
```diff
@@ -0,0 +1,94 @@
...
+        A list of Tamr backups
+    """
+    url = URL(instance=instance, path='backups')
+    r = session.get(str(url))
+    if r.status_code == 404:
+        raise NotFound(str(url))
+    return response.successful(r).json()
+
+
+def get_by_id(session: Session, instance: Instance, backup_id: str) -> dict:
```
skalish, line 37: To mirror the `from_resource_id` functions for projects/datasets/etc, this could be `from_id`.

tamr_client/instance/backup.py
abafzal: Canceling an already canceled backup: ```
{
  "status": 400,
  "class": "javax.ws.rs.BadRequestException",
  "message": "Backup has already completed with status CANCELED. Cancellation failed!",
```

Canceling a failed backup: ```
{
  "status": 400,
  "class": "javax.ws.rs.BadRequestException",
  "message": "Backup has already completed with status FAILED. Cancellation failed!",
```

Canceling a successful/finished backup: ```
{
  "status": 400,
  "class": "javax.ws.rs.BadRequestException",
  "message": "Backup has already completed with status SUCCEEDED. Cancellation failed!",
```

Canceling a nonexisting backup: ```
{
  "status": 404,
  "class": "com.tamr.common.except.ResourceNotFoundException",
  "message": "Invalid backup ID: asfkjhasdlkfjahsd",
```

tamr_client/instance/backup.py
abafzal: it would be `tc.instance.backup.backups()` which i am fine with

tamr_client/instance/backup.py
abafzal: if we have `backups()` to get all backups shouldn't it be something like `backup_by_id()`. It seems inconsistent to have the word backup in one function name but not the other.

tamr_client/instance/backup.py
skalish: I think it's preferable to take advantage of the namespace and avoid repetition.  Any function in the `backup` module, probably shouldn't have the word backup in its name.  My early suggestion would involve moving this function `backups` into `instance.py` but I definitely see the weirdness in separating it from the other functions.

tamr_client/instance/backup.py
skalish: This naming definitely will be influenced by the decision on the function that retrieves all backups.  See the comment there about avoiding repetition in the names.

tamr_client/instance/backup.py
abafzal: what about `tc.instance.backup.all()` and `tc.instance.backup.from_id()`

skalish: That seems pretty reasonable to me 👍

tamr_client/instance/__init__.py
```diff
@@ -0,0 +1,3 @@
+from tamr_client.instance import _instance, backup
```
pcattori, line 1: Don't expose non-intended interfaces:

```suggestion
from tamr_client.instance import backup
```

tamr_client/instance/backup.py
pcattori, line 20: We should model backups as a type.

Also: `list` is the list constructor in Python, but not a valid type for type-checking. Use `typing.List` instead.

```suggestion
def get(session: Session, instance: Instance) -> List[Backup]:
```

tamr_client/instance/backup.py
pcattori: Should be named `from_resource_id`, since `id` could be confused with the universal ID (e.g. https://github.com/Datatamer/tamr-client/pull/438/files#diff-362c9b0ea842fd1de6fde077c5610098R10 )

tamr_client/instance/backup.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Returns:
+        Information on initiated backup
+    """
+    url = URL(instance=instance, path="backups")
+    r = session.post(str(url))
+    return response.successful(r).json()
+
+
+def cancel(session: Session, instance: Instance, backup_id: str) -> dict:
```
pcattori, line 73: After we model backups as types, we should be operating on that type:

```suggestion
def cancel(session: Session, instance: Instance, backup: Backup) -> dict:
```

tamr_client/instance/backup.py
skalish: To conform to the language in those messages, maybe `AlreadyCompleted` and pass the message.  I'm not totally sure.

abafzal: Maybe something like `InvalidOperation`? This way it doesn't have to be specific to the cancel task, it can be applied to other types of operations we may run across that are invalid/not possible for other reasons.

tamr_client/instance/backup.py
abafzal: updated to `from_resource_id` for now. Although I kind of like `by_resource_id` better when fetching from Tamr

tamr_client/instance/backup.py
abafzal: using `get_all()` instead to keep consistent with transformations.py

tamr_client/_types/backup.py
```diff
@@ -0,0 +1,24 @@
...
+        path
+        state
+        errorMessage
+    """
+
+    url: URL
+    resource_id: str
+    path: str
+    state: str
+    errorMessage: str
```
pcattori, line 24: camelCase -> snake_case:

```suggestion
    error_message: str
```

tamr_client/backup.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Returns:
+        Initiated backup
+    """
+    url = URL(instance=instance, path="backups")
+    r = session.post(str(url))
+    return response.successful(r).json()
+
+
+def cancel(session: Session, instance: Instance, backup: Backup) -> dict:
```
skalish, line 91:
```suggestion
def cancel(session: Session, instance: Instance, backup: Backup) -> JsonDict:
```
You can find this in `_types.json`

tamr_client/backup.py
```diff
@@ -0,0 +1,120 @@
...
+        A list of Tamr backups
+
+    Raises:
+        backup.NotFound: If no backup found at the specified URL
+    """
+    url = URL(instance=instance, path="backups")
+    r = session.get(str(url))
+    if r.status_code == 404:
+        raise NotFound(str(url))
+    backups = list(_from_json(i) for i in response.successful(r).json())
```
pcattori, line 54: nit: if already using a `for`-loop expression, list comprehension is more idiomatic than wrapping in `list(...)`:

```suggestion
    backups = [_from_json(i) for i in response.successful(r).json()]
```

tamr_client/backup.py
```diff
@@ -0,0 +1,120 @@
...
+        backup.InvalidOperation: If attempting an invalid operation
+    """
+    url = URL(instance=instance, path="backups")
+    r = session.post(str(url))
+    if r.status_code == 400:
+        raise InvalidOperation(str(url), r.json()["message"])
+    return _from_json(response.successful(r).json())
+
+
+def cancel(session: Session, instance: Instance, backup: Backup) -> Backup:
```
pcattori, line 99: Instance should be inferred from the `backup`'s URL.

```suggestion
def cancel(session: Session, backup: Backup) -> Backup:
```

tamr_client/_types/backup.py
```diff
@@ -0,0 +1,24 @@
+from dataclasses import dataclass
+
+from tamr_client._types.url import URL
+
+
+@dataclass(frozen=True)
+class Backup:
```
pcattori, line 7: Do we need a better name since `Backup` here refers to the operation, not the resulting .zip artifact?

tamr_client/backup.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _from_json(data: JsonDict) -> Backup:
+    """Make backup from JSON data (deserialize).
+
+    Args:
+        data: Backup JSON data from Tamr server
+    """
+    return Backup(
+        url=data["id"],
```
pcattori, line 29: The universal ID (the one starting with `unify://`), accessed here via `data["id"]`, is not a valid HTTP URL that we can query.

To remedy, the URL should be an argument to `_from_json` passed down by the caller.

tamr_client/_types/backup.py
abafzal: Tamr refers to these (the JSON objects that contain information about a backup operation) as backups so i think its fine to leave as is

pcattori: Can you add to the docstring saying that?

tamr_client/backup.py
abafzal: changed

tamr_client/backup.py
```diff
@@ -0,0 +1,120 @@
...
+        resource_id=data["relativeId"],
+        path=data["backupPath"],
+        state=data["state"],
+        error_message=data["errorMessage"],
+    )
+
+
+def get_all(session: Session, instance: Instance) -> List[Backup]:
+    """Get information on Tamr backups.
```
skalish, line 38: I think this line in the docstring (and in `from_resource_id`) could be made more specific.  If we are using the terminology `Backup` for the job/process/operation that backs up an instance of Tamr, this should instead say
```python
Get all backups that have been initiated for a Tamr instance
```

skalish:
Would it make sense to include a `poll` function here? Backups are like operations in that their state is expected to change and a common task is checking if a backup is finished.  I don't think I would go as far as creating a `wait` function or anything else synchronous, but that is an additional possibility.

tamr_client/_types/backup.py
```diff
@@ -0,0 +1,24 @@
...
+    Args:
+        url
+        resource_id
+        path
+        state
+        error_message
+    """
+
+    url: URL
+    resource_id: str
```
pcattori, line 21: I don't think we want to model `resource_id` explicitly, but rather rely on the `url` field instead.

```suggestion
```

Note: should be removed from docstring as well.

tamr_client/_types/restore.py
```diff
@@ -0,0 +1,24 @@
...
+    Args:
+        url
+        resource_id
+        backup_path
+        state
+        error_message
+    """
+
+    url: URL
+    resource_id: str
```
pcattori, line 21: Same comment about `resource_id`:

```suggestion
```

tamr_client/backup.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Raises:
+        backup.NotFound: If no backup found at the specified URL
+        backup.InvalidOperation: If attempting an invalid operation
+    """
+    r = session.post(f"{backup.url}:cancel")
+    if r.status_code == 404:
+        raise NotFound(f"{backup.url}:cancel")
+    if r.status_code == 400:
+        raise InvalidOperation(f"{backup.url}:cancel", r.json()["message"])
```
pcattori, line 126: nit: Could pull out `f"{backup.url}:cancel"`:

```suggestion
    cancel_url = f"{backup.url}:cancel"
    r = session.post(cancel_url)
    if r.status_code == 404:
        raise NotFound(cancel_url)
    if r.status_code == 400:
        raise InvalidOperation(cancel_url, r.json()["message"])
```

tamr_client/restore.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Raises:
+        restore.NotFound: If no backup file found at the specified path
+        restore.InvalidOperation: If attempting an invalid operation
+    """
+    r = session.post(f"{restore.url}:cancel")
+    if r.status_code == 404:
+        raise NotFound(str(restore.url))
+    if r.status_code == 400:
+        raise InvalidOperation(str(restore.url), r.json()["message"])
```
pcattori, line 95: Should the exceptions be pointing to `restore.url`? or `f"{restore.url}:cancel"`?

tests/tamr_client/test_backup.py
```diff
@@ -0,0 +1,48 @@
...
+        "stage": "",
+        "errorMessage": "",
+        "created": "2020-08-17_21-32-10-961",
+        "lastModified": "2020-08-17_21-51-57-600",
+    }
+    backup = tc.backup._from_json(
+        url=tc.URL(path="backups/2020-08-17_21-32-10-961"), data=data
+    )
+
+    tc.backup.cancel(session=s, backup=backup)
```
pcattori, line 48: Is this trying to cancel an already canceled backup?

pcattori (changes requested):

tests/tamr_client/test_backup.py
skalish: Yeah, I think this is just a confusing object to start with.  We don't do any checking of the state of the backup, so it is functionally the same, but I think it makes sense to change `"CANCELED"` to `"RUNNING"` here.

tamr_client/restore.py
skalish: For `InvalidOperation` it should certainly be `f"{restore.url}:cancel"`.  For `NotFound` I think it could be either, but based on what is done in `backup.py` I would do the same here (`f"{restore.url}:cancel"` again).

tamr_client/_types/backup.py
skalish: I think this is reasonable to avoid redundancy (and possible conflicts in user-constructed objects).  This does mean it is moderately harder to get an updated version of a `Backup`, since you can no longer use `backup = tc.backup.by_resource_id(s, instance, backup.resource_id)`.  It isn't much harder to do `str(backup.url).split("/")[-1]`, but maybe this change justifies adding a `poll` function for this purpose.

pcattori: Rather than the error-prone `resource_id` field, we could either provider a `resource_id(backup)`. But semantically, you should be able to just pass in the `backup` to a function to get what you want.

I.e. why are you even getting the resource ID? To poll the backup? Then just call `poll(backup)`.

So yes, I think this justifies a `poll` function 😄

pcattori (approved):
LGTM

I actually don't think we need to docstring for `Backup` since we link to the Tamr docs.

skalish:
> LGTM
> 
> I actually don't think we need to docstring for `Backup` since we link to the Tamr docs.

Agreed. I think this is all set.

PR #440, opened 2020-08-19
TC: Add functions for basic mastering workflow operations

skalish:
# ↪️ Pull Request

This PR adds functions to `tamr_client` that are needed to bring a Tamr Mastering project up to date (but not make changes).  

Function names are intended to mirror the language used in the UI, but also reflect the separate underlying (versioned) API calls.  For instance:
1. "Update Results" in the UI ~ running `update_pair_results` + `update_high_impact_pairs` + `update_clusters`.
2. "Apply Feedback and Update Results" ~ `apply_feedback` + (the above functions)

## 💻 Examples

```python
import tamr_client as tc

s = tc.Session(...)
project = tc.MasteringProject(...)

# Run mastering jobs
tc.mastering.update_unified_dataset(s, project)

tc.mastering.estimate_pairs(s, project)           # optional 
tc.mastering.generate_pairs(s, project)

tc.mastering.apply_feedback(s, project)           # optional (trains model)
tc.mastering.update_pair_results(s, project)
tc.mastering.update_high_impact_pairs(s, project) # optional
tc.mastering.update_cluster_results(s, project)

tc.mastering.publish_clusters(s, project)
```

## ✔️ PR Todo

### DIFF
--- a/tamr_client/mastering/_mastering.py
+++ b/tamr_client/mastering/_mastering.py
@@ -1,3 +1,9 @@
+"""
+Tamr - Mastering
+See https://docs.tamr.com/docs/overall-workflow-mastering
+
+The terminology used here is consistent with Tamr UI terminology
+"""
 from tamr_client import operation
 from tamr_client._types import MasteringProject, Operation, Session
 from tamr_client.dataset import unified
@@ -44,7 +50,7 @@ def apply_feedback(session: Session, project: MasteringProject) -> Operation:
 
 
 def update_pair_results(session: Session, project: MasteringProject) -> Operation:
-    """Update record pair predictions according to the latest pair-matching model
+    """Updates record pair predictions according to the latest pair-matching model
 
     Args:
         project: Tamr Mastering project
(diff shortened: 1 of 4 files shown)


abafzal:
Looks good to me but would like to see tests for each new function.

pcattori:
I'm actually not sure that I agree about testing (if we are talking about unit tests); unit tests would just check that the correct endpoints are getting called.

We should add integration or E2E tests that exercise these functions instead.

tamr_client/mastering/_mastering.py
```diff
@@ -0,0 +1,83 @@
+from tamr_client import operation
```
pcattori, line 1: Can we include a module-level docstring that:
- points to the Mastering docs in docs.tamr.com
- states that the terminology used here is consistent with UI terminology

skalish: There is a docstring in `mastering/__init__.py` that satisfied part of this:
```python
"""
Tamr - Mastering
See https://docs.tamr.com/docs/overall-workflow-mastering
"""
```
Is it more appropriate for the module-level docstring to live in `_mastering.py`, and if so should there be any docstring in `__init__.py`?

pcattori: I think the duplication is ok since it's only 1-2 lines of docstring duplication. I think contributors will benefit from having the links in both places.

PR #445, opened 2020-08-21
TC: by_name functions for projects and datasets

skalish:
# ↪️ Pull Request

Adds functions for getting Tamr projects and datasets by name

## 💻 Examples

```python
import tamr_client as tc

auth = tc.UsernamePasswordAuth(...)
s = tc.session.from_auth(auth)
instance = tc.Instance("http", "localhost", 9100)

dataset = tc.dataset.by_name(s, instance, "My Dataset")
project = tc.project.by_name(s, instance, "My Project")
```

## ✔️ PR Todo


skalish:
In this PR, there is a language mismatch between the `from_resource_id` and `by_name` functions.  I actually think "by" makes more sense when getting a resource from Tamr, whereas "from" fits de-serialization. So I might advocate for `by_name`, `by_resource_id`, and even `_by_url`, leaving only `_from_json` (and of course functions like `unified.from_project`).  This language is already being used in the docstrings for these functions: "Get dataset by URL" vs "Make dataset from JSON data (deserialize)".

ianbakst:
I agree with this name change. This preposition makes more sense to me.

skalish:
I've included the `from_resource_id` -> `by_resource_id` and `_from_url` -> `_by_url` renaming at the end of this PR, but I'm happy to roll back one or both if we decided to keep the old naming.

tests/tamr_client/fake_json/dataset/test_dataset/test_by_name.json
```diff
@@ -0,0 +1,36 @@
+[
+    {
+        "request": {
+            "method": "GET",
+            "path": "datasets?filter=name==dataset 1 name"
```
pcattori, line 5: nit: Ideally, tests should be as realistic as possible, so `dataset 1 name` should be replaced with something like `Customers_2019.csv` or something. We are inconsistent about this and I think we can wait to fix this until when we switch over to faking a test server.

PR #452, opened 2020-09-02
Dataset CRUD + Project List Functions

skalish:
# ↪️ Pull Request

This PR closes the feature gap for Tamr dataset creation and deletion.  Additionally, it adds functionality to get the list of all datasets/projects on an instance (as `Dataset`/`Project` objects), optionally passing a filter.  The functionality to truncate a dataset (delete all records) is added via `tc.record.delete_all`.

## 💻 Examples

```python
import tamr_client as tc

session = ...
instance = ...

# Make a new dataset
new_dataset = tc.dataset.create(
    s,
    instance,
    name="My New Dataset",
    key_attribute_names=("primary_key",),
    description="A new dataset created with TC",
)
# Delete the dataset
tc.dataset.delete(s, new_dataset)

# Get all datasets
dataset_list = tc.dataset.get_all(s, instance)

# Get all projects
project_list = tc.project.get_all(s, instance)

# Get all datasets that match a filter
wobbly_dataset_list = tc.dataset.get_all(s, instance, filter="name==wobbly")

# Truncate records in a dataset
dataset = ...
tc.record.delete_all(s, dataset)
```

## ✔️ PR Todo

### DIFF
--- a/tamr_client/dataset/_dataset.py
+++ b/tamr_client/dataset/_dataset.py
@@ -213,3 +213,44 @@ def get_all(
         dataset = _from_json(dataset_url, dataset_json)
         datasets.append(dataset)
     return tuple(datasets)
+
+
+def create(
+    session: Session,
+    instance: Instance,
+    *,
+    name: str,
+    key_attribute_names: Tuple[str, ...],
+    description: Optional[str] = None,
+    external_id: Optional[str] = None,
+) -> Dataset:
+    """Create a dataset in Tamr.
+
+    Args:
+        instance: Tamr instance
+        name: Dataset name
+        key_attribute_names: Dataset primary key attribute names
+        description: Dataset description
+        external_id: External ID of the dataset
+
+    Returns:
+        Dataset created in Tamr
+
+    Raises:
+        requests.HTTPError: If any other HTTP error is encountered.
+    """
+    data = {
+        "name": name,
+        "keyAttributeNames": key_attribute_names,
+        "description": description,
+        "externalId": external_id,
+    }
+
+    dataset_url = URL(instance=instance, path="datasets")
+    r = session.post(url=str(dataset_url), json=data)
+
+    data = response.successful(r).json()
+    dataset_path = data["relativeId"]
+    dataset_url = URL(instance=instance, path=str(dataset_path))
+
+    return _by_url(session=session, url=dataset_url)
--- a/tamr_client/dataset/_dataset.py
+++ b/tamr_client/dataset/_dataset.py
@@ -3,7 +3,7 @@ See https://docs.tamr.com/reference/dataset-models
 """
 from copy import deepcopy
 from dataclasses import replace
-from typing import Tuple
+from typing import List, Optional, Tuple, Union
 
 from tamr_client import operation, response
 from tamr_client._types import (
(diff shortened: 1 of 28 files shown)


skalish:
The `POST /v1/datasets` creation API call returns a status `400` when trying to create a dataset with a name that is already in use, which makes handling this with an informative `dataset.AlreadyExists` exception difficult.

skalish:
I modeled the `dataset.create` function on the similar `attribute.create` function in that

1. The dataset is created in Tamr by calling `POST /v1/datasets`
2. The `Dataset` object is created by calling `GET /v1/datasets/{datasetId}` and deserializing the result

In both cases, I think the response from 1 contains the JSON necessary to create the `Dataset`.  Is there value in making two calls rather one here?  Maybe in some sense ensuring that the new dataset can be retrieved with `GET`.

tamr_client/dataset/_dataset.py
```diff
@@ -157,3 +157,100 @@ def materialize(session: Session, dataset: Dataset) -> Operation:
...
+    data = {
+        "name": name,
+        "keyAttributeNames": key_attribute_names,
+        "description": description,
+        "externalId": external_id,
+    }
+
+    dataset_url = URL(instance=instance, path="datasets")
+    r = session.post(url=str(dataset_url), json=data)
+
```
pcattori, line 251: For handling "already exists" errors, we can try to match on the response text to distinguish that error from other 400 errors:

```suggestion
if r.status_code == 400 and f'Dataset "{name}" already exists' in r.text:
    raise AlreadyExists(name)
```

Even if the server changes its error codes for "already exists" upstream, we will still at least catch any errors via `response.successful` on the next line.

pcattori (changes requested):

tests/tamr_client/fake_json/dataset/test_dataset/test_create_dataset_already_exists.json
```diff
@@ -0,0 +1,22 @@
...
+                    "primary_key"
+                ],
+                "description": "a new dataset",
+                "externalId": null
+            }
+        },
+        "response": {
+            "status": 400,
+            "json": {
+                "message": "Dataset \"tc test dataset\" already exists"
```
pcattori, line 18: nit: should this be:

```suggestion
                "message": "Dataset \"new dataset\" already exists"
```

pcattori (approved):
LGTM

Minor nit about test data

tests/tamr_client/fake_json/dataset/test_dataset/test_create_dataset_already_exists.json
skalish: Will change, just a minute.

skalish:
> I modeled the `dataset.create` function on the similar `attribute.create` function in that
> 
> 1. The dataset is created in Tamr by calling `POST /v1/datasets`
> 2. The `Dataset` object is created by calling `GET /v1/datasets/{datasetId}` and deserializing the result
> 
> In both cases, I think the response from 1 contains the JSON necessary to create the `Dataset`. Is there value in making two calls rather one here? Maybe in some sense ensuring that the new dataset can be retrieved with `GET`.

@pcattori any thoughts on this before merging?

pcattori:
> > I modeled the `dataset.create` function on the similar `attribute.create` function in that
> > 
> > 1. The dataset is created in Tamr by calling `POST /v1/datasets`
> > 2. The `Dataset` object is created by calling `GET /v1/datasets/{datasetId}` and deserializing the result
> > 
> > In both cases, I think the response from 1 contains the JSON necessary to create the `Dataset`. Is there value in making two calls rather one here? Maybe in some sense ensuring that the new dataset can be retrieved with `GET`.
> 
> @pcattori any thoughts on this before merging?

I think its conceptually simpler to always get the data via GET. If we run into perf issues, we can always change this since its an implementation detail not exposed to the end user.

Issue #455, opened 2020-09-14
Remove `simplejson` as a dependency


skalish:
# 💬 RFC

`simplejson` is now the only dependency in this project whose build is machine-dependent (i.e. if you `pip download` on MacOS, the result cannot be used on Ubuntu).  Additionally, to my knowledge it is only used to a very limited extent: to allow a toggle to `ignore_nan` values in the record update operation of `tamr_unify_client`. This means a record (as dictionary) containing at least one `NaN` that is upserted to Tamr with the client will cause an `HTTPError` if `ignore_nan == True` or will have this value coerced to `null` if `ignore_nan == False`.

1. This dependency is not present in `tamr_client`, however, there are no plans to remove `tamr_unify_client` even after the official promotion of the BETA to the primary implementation.  For this reason, I think it is important that `simplejson` be removed to ease the installation process.

2. If it is decided that we remove this, the `ignore_nan` functionality needs to either be added in its place or be dropped as an unneeded feature.  In the former case, it might also be prudent to add similar functionality to `tamr_client`.

skalish:
Regarding the suggestion to allow flexibility in passing other `json_args` https://github.com/Datatamer/tamr-client/pull/209#discussion_r307306802: while this is true, I am unaware of any cases where this functionality is being taken advantage of, while I have seen at least two cases where getting `tamr-unify-client` installed on a customer machine without access to the internet was quite painful.

pcattori:
If the main issue is that including `NaN` will silently fail or fail downstream, we could use the [allow_nan](https://docs.python.org/3/library/json.html#basic-usage) parameter (setting `allow_nan=False`) to fail fast.

Then users can fix their records before passing them to TC functions with `allow_nan=False`.

skalish:
The original issue #194 was about including the option to ignore `NaN`s, that is, having `tamr_client` take care of the conversion from `NaN` to `null` if the user chooses.  The `allow_nan=False` solution does avoid the original problematic behavior of downstream 400 errors, but this should be recognized as an intentional decision to no longer provide `NaN`->`null` functionality (which is relied upon by some field code, though I am unsure if any of it is still in use).

PR #457, opened 2020-09-17
No simplejson

skalish:
# ↪️ Pull Request

Relates to #455 

This PR explores two ways that `simplejson` could be removed from the dependency list, as described in #455.  The two commits are not really consecutive, but are instead alternatives.

1. The first commit removes `simplejson` via a breaking change, where the `_update_records`/`upsert_records` methods no longer have the `ignore_nan` parameter.  This means when using these functions, it is on the user to make sure `NaN`s are not present in the records being uploaded, though the function does use `json`'s `allow_nan=False` option to raise an error before making the destined-to-fail request to the server if `NaN`'s are present.  Handling of `NaN`s **is** present in the `upsert_from_dataframe` function, since this is the primary case where such behavior is desired.  This is done in the same was as in `tamr_client` and both does not depend on `simplejson` and is robust.

2. The second commit also removes `simplejson`, but leaves the `ignore_nan` parameter in `_update_records`/`upsert_records`.  This parameter is removed from the `upsert_from_dataframe` and `create_from_dataframe` functions, since these are changed to use the `tamr_client` DataFrame solution.  The custom implementation of `ignore_nan` for `_update_records`/`upsert_records` is simple, but limited.  While serializing JSON of each update, `allow_nan=False` is used and raised errors are handled by looking for Python `NaN` values at the **top-level** within a record.  This means that the breaking change is limited to the last of the following example updates:
```python
{
  "action": "CREATE", 
  "recordId": 1,
  "record": {"pk": 1, "attr": 42}  # Nothing to catch
}
{
  "action": "CREATE",
  "recordId": 2,
  "record": {"pk": 2, "attr": NaN}  # Caught and converted
}
{
  "action": "CREATE",
  "recordId": 2,
  "record": {"pk": 2, "attr": {"a": NaN, "b": 42}}  # Not converted, will raise ValueError
}
```  

## ✔️ PR Todo

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -446,14 +446,6 @@ six = "*"
 tests = ["pytest", "coverage (>=3.7.1,<5.0.0)", "pytest-cov", "pytest-localserver", "flake8"]
 
 [[package]]
-category = "main"
-description = "Simple, fast, extensible JSON encoder/decoder for Python"
-name = "simplejson"
-optional = false
-python-versions = ">=2.5, !=3.0.*, !=3.1.*, !=3.2.*"
-version = "3.16.0"
-
-[[package]]
 category = "dev"
 description = "Python 2 and 3 compatibility utilities"
 name = "six"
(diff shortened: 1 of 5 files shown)


skalish:
This PR is superseded by #468

PR #468, opened 2020-10-25
Remove dependency on simplejson and use a custom encoder when needed

skalish:
# ↪️ Pull Request

This PR breaks the dependency on `simplejson` in two ways:
### `tamr_client`
The onus is placed on the user to ensure that their record updates do not contain `NaN` values, and that these are instead cast to the Python `None`.  `NaN` values in a `pandas` DataFrame will not cause issues in the `dataframe.upsert` function since the generated dictionaries are sanitized by the Series method `to_json`.

### `tamr_unify_client`
The usage of `simplejson` in the method `Dataset._update_records` has been replaced by using a custom JSON encoder when passing `ignore_nan=True`.  This is now the only `json_arg` that is supported.

The custom encoder, defined in `_custom_encoder.py` is minimally changed from the standard library encoder to cast `NaN` (and `Infinity`/`-Infinity`) to `'null'` in the style of `simplejson`.  Compare to [json](https://github.com/python/cpython/blob/83c86cf54b36a7325f615f5adf22b28e48f0e72d/Lib/json/encoder.py#L223-L243) and [simplejson](https://github.com/simplejson/simplejson/blob/c51668542b406c5aadb052a1513245307cc8b014/simplejson/encoder.py#L328-L353).

The method `Dataset.upsert_from_dataframe` has had the argument `ignore_nan` removed, since this is the only reasonable behavior.  The implementation there has been changed to mirror the solution in `tamr-client`. (Maybe the parameter should be left to avoid a breaking change and be raised as a deprecation warning).

## ✔️ PR Todo


tamr_unify_client/_custom_encoder.py
```diff
@@ -0,0 +1,457 @@
+# flake8: noqa: C901
+"""Implementation of JSONEncoder
+"""
```
pcattori, line 3: Could you replace "Implementation of JSONEncoder" with a module-level docstrings that links to where this is adapted from? Ideally, you could also call out in the module docstring which parts of this are modified from the source.

tamr_unify_client/dataset/resource.py
```diff
@@ -64,20 +64,21 @@ def attributes(self):
         alias = self.api_path + "/attributes"
         return AttributeCollection(self.client, alias)
 
-    def _update_records(self, updates, **json_args):
+    def _update_records(self, updates, ignore_nan=False):
```
skalish, line 67: Alternatively, we could leave `**json_args`, pop `ignore_nan` from the resulting dictionary, and allow the remainder to be passed to the standard library `json.dumps`, where an error will be raised for invalid kwargs.

tamr_unify_client/dataset/resource.py
```diff
@@ -64,20 +64,21 @@ def attributes(self):
...
         """Send a batch of record creations/updates/deletions to this dataset.
         You probably want to use :func:`~tamr_unify_client.dataset.resource.Dataset.upsert_records`
         or :func:`~tamr_unify_client.dataset.resource.Dataset.delete_records` instead.
 
         :param records: Each record should be formatted as specified in the `Public Docs for Dataset updates <https://docs.tamr.com/reference#modify-a-datasets-records>`_.
         :type records: iterable[dict]
-        :param `**json_args`: Arguments to pass to the JSON `dumps` function, as documented `here <https://simplejson.readthedocs.io/en/latest/#simplejson.dumps>`_.
-            Some of these, such as `indent`, may not work with Tamr.
+        :param ignore_nan: Whether to treat `NaN` values as null.  Unconverted `NaN`s will raise an error if found.
```
pcattori, line 74: Should we tell users to stop relying on `ignore_nan`? Wondering if we want to deprecate and later remove this parameter.

Alternatively, we could simply say that TUC itself is now deprecated.

tamr_unify_client/dataset/resource.py
```diff
@@ -98,7 +99,7 @@ def upsert_from_dataframe(
         Args:
             df: The data to upsert records from.
             primary_key_name: The name of the primary key of the dataset.  Must be a column of `df`.
-            ignore_nan: Whether to convert `NaN` values to `null` before upserting records to Tamr. If `False` and `NaN` is in `df`, this function will fail. Optional, default is `True`.
+            ignore_nan: Legacy parameter that does nothing
```
pcattori, line 102: Same question about deprecation/later removal. Should we print a deprecation warning when people use this parameter?

pcattori (changes requested):

tamr_unify_client/_custom_encoder.py
skalish: How's it look now?

tamr_unify_client/dataset/resource.py
pcattori: I think being aggressive about removing these kwargs is a good thing. We have reasonable confidence that that only kwarg being used is `ignore_nan`. Worst-case, we can do a patch release that does what you suggest^ (pop `ignore_nan`, pass the rest to `json.dumps`) if we find out there are use-cases that use other kwargs.

tamr_unify_client/dataset/resource.py
pcattori: Something as simple as "WARN: 'ignore_nan' is deprecated. Users are expected to provide valid JSON representations instead" or something like that.

tamr_unify_client/_custom_encoder.py
```diff
@@ -0,0 +1,459 @@
...
+                return encode_basestring(o)
+        # This doesn't pass the iterator directly to ''.join() because the
+        # exceptions aren't as detailed.  The list call should be roughly
+        # equivalent to the PySequence_Fast that ''.join() would do.
+        chunks = self.iterencode(o, _one_shot=True)
+        if not isinstance(chunks, (list, tuple)):
+            chunks = list(chunks)
+        return "".join(chunks)
+
+    def iterencode(self, o, _one_shot=False):
```
pcattori, line 204: I'm wondering if we can just make a new JSON encoder class that inherits from `json.JSONEncoder` and only overrides `iterencode` / `floatstr`. That would mean we don't have to copy the entire `JSONEncoder` implementation

skalish: Good idea.  I gave this a shot, but not 100% confident it's correct.

tamr_unify_client/_custom_encoder.py
```diff
@@ -0,0 +1,89 @@
...
+    from _json import make_encoder as c_make_encoder
+except ImportError:
+    c_make_encoder = None
+
+INFINITY = float("inf")
+encode_basestring = c_encode_basestring or py_encode_basestring
+encode_basestring_ascii = c_encode_basestring_ascii or py_encode_basestring_ascii
+
+
+class CustomEncoder(JSONEncoder):
```
pcattori, line 31: Instead of `CustomEncoder` how about something like `IgnoreNanEncoder`?

```suggestion
class IgnoreNanEncoder(JSONEncoder):
```

pcattori (changes requested):

skalish:
Just as a note, testing on this topic is insufficient at the moment.

tamr_unify_client/_custom_encoder.py
skalish: Done.

tamr_unify_client/_custom_encoder.py
skalish: I think it's necessary to override `iterencode` and not just `floatstr` but this is still a major improvement.

tamr_unify_client/dataset/resource.py
skalish: Any preference between `DeprecationWarning` and `PendingDeprecationWarning`?

pcattori: `DeprecationWarning` is what we want since there's already viable alternatives.

pcattori (approved):
LGTM, just missing the changelog

tamr_unify_client/_custom_encoder.py
```diff
@@ -0,0 +1,67 @@
+"""Adaptation of the Python standard library JSONEncoder to encode `NaN` as 'null'
```
pcattori, line 1: Should we rename this file to `_ignore_nan_encoder.py`?

skalish: Done

skalish:
In terms of changelog, I think there is a breaking change in that the only json-encoding parameter that is accepted now is `ignore_nan`.  Besides that, what is the best way to communicate these changes that are only user-facing in that `simplejson` is no longer a dependency, and deprecation warning are now added?

PR #475, opened 2020-11-24
Update checkout and setup-python actions to v2

skalish:
Github actions checkout v1 and setup-python v1.1.1 have been officially deprecated as of last week. I believe this is the simple fix needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,9 @@ jobs:
   Lint:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: Install python
-      uses: actions/setup-python@v1.1.1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.6
     - name: Install poetry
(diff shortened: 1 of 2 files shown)


skalish:
My only uncertainty here is if `dschep/install-poetry-action@v1.2` needs to be changed as well in `release.yml`.

pcattori:
> My only uncertainty here is if `dschep/install-poetry-action@v1.2` needs to be changed as well in `release.yml`.

Looks like [dschep/install-poetry-action@v1.2 is deprecated](https://github.com/dschep/install-poetry-action). That repo recommends [snok/install-poetry](https://github.com/snok/install-poetry), but I've also seen [abatilo/actions-poetry](https://github.com/abatilo/actions-poetry).

pcattori (approved):
LGTM

We can either address the release workflow in this PR or in a follow-on PR.

skalish:
> LGTM
> 
> We can either address the release workflow in this PR or in a follow-on PR.

I'm unsure of how to test the release workflow and the PR issue is blocking, so I agree on doing a follow-on.

PR #478, opened 2021-01-14
TC: Create dataset from DataFrame

skalish:
# ↪️ Pull Request

This PR adds functionality to `tamr_client` to allow the creation of a Tamr dataset from a `pandas` DataFrame, including:
- dataset creation
- attribute creation with all attributes created as type array(string)
- upload of all rows of the DataFrame as records

Related to #247 

## 💻 Examples
```
import pandas as pd
import tamr_client as tc

instance = tc.Instance(host='10.10.0.121')

df = pd.DataFrame({"pk": [1, 2], "val": [11, 12]})

new_dataset = tc.dataframe.create(s, instance, df, name="my_dataset", primary_key_name="pk")
```

## ✔️ PR Todo

- [x] Testing for this change

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,3 +39,4 @@ For more see the [official docs](https://tamr-client.readthedocs.io/en/stable/).
 ## Maintainers
 
 - [Pedro Cattori](https://github.com/pcattori)
+- [Samuel Kalish](https://github.com/skalish)
--- a/tamr_client/dataset/dataframe.py
+++ b/tamr_client/dataset/dataframe.py
@@ -6,13 +6,23 @@ import json
 import os
 from typing import Optional, TYPE_CHECKING
 
-from tamr_client import primary_key
-from tamr_client._types import Dataset, JsonDict, Session
+import requests
+
+from tamr_client import attribute, dataset, primary_key
+from tamr_client._types import Dataset, Instance, JsonDict, Session
 from tamr_client.dataset import record
+from tamr_client.exception import TamrClientException
 
 BUILDING_DOCS = os.environ.get("TAMR_CLIENT_DOCS") == "1"
 if TYPE_CHECKING or BUILDING_DOCS:
     import pandas as pd
+
+
+class CreationFailure(TamrClientException):
+    """Raised when a dataset could not be created from a pandas DataFrame
+    """
+
+    pass
 
 
 def upsert(
@@ -27,7 +37,8 @@ def upsert(
     Args:
         dataset: Dataset to receive record updates
         df: The DataFrame containing records to be upserted
-        primary_key_name: The primary key of the dataset.  Must be a column of `df`. By default the key_attribute_name of dataset
+        primary_key_name: The primary key of the dataset.  Must be a column of `df`. By default the
+            key_attribute_name of dataset
 
     Returns:
         JSON response body from the server
@@ -41,14 +52,7 @@ def upsert(
         primary_key_name = dataset.key_attribute_names[0]
 
     # preconditions
-    if primary_key_name in df.columns and primary_key_name == df.index.name:
-        raise primary_key.Ambiguous(
-            f"Index {primary_key_name} has the same name as column {primary_key_name}"
-        )
-    elif primary_key_name not in df.columns and primary_key_name != df.index.name:
-        raise primary_key.NotFound(
-            f"Primary key: {primary_key_name} is not DataFrame index name: {df.index.name} or in DataFrame column names: {df.columns}"
-        )
+    _check_primary_key(df, primary_key_name)
 
     # promote primary key column to index
     if primary_key_name in df.columns:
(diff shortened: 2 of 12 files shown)


skalish:
This still needs to be tested against a real instance.  After that, will mark as ready for review.

yizhiyintamr:
one questions about this example: Does all values need to be in an array even if a column only has a single value `{"col1":["val1'],"col2":["val1","val2"]}`?

skalish:
> one questions about this example: Does all values need to be in an array even if a column only has a single value `{"col1":["val1"],"col2":["val1","val2"]}`?
> I think I was confused about the construction of data frame. I think this example looks good. :)

It looks like construction of a DataFrame from a dictionary of columns requires the arrays to be the same length, so nulls should be passed in explicitly, e.g. `{"col1": ["val1", None], "col2": ["val1", "val2"]}`.  FYI, in contrast to my example, our best practices suggestions for using `pandas` with the client are here: https://tamr-client.readthedocs.io/en/stable/user-guide/pandas.html (and improvements are always welcome)

README.md
```diff
@@ -4,7 +4,7 @@ Programmatically 💻 interact with Tamr using Python 🐍
 [![Version](https://img.shields.io/pypi/v/tamr-unify-client.svg?style=flat-square)](https://pypi.org/project/tamr-unify-client/)
 [![Documentation Status](https://readthedocs.org/projects/tamr-client/badge/?version=stable&style=flat-square)](https://tamr-client.readthedocs.io/en/stable/?badge=stable)
 [![Build Status](https://img.shields.io/github/workflow/status/Datatamer/tamr-client/CI?&style=flat-square)](https://github.com/Datatamer/tamr-client/actions?query=workflow%3ACI)
-![Supported Python Versions](https://img.shields.io/pypi/pyversions/tamr-unify-client.svg?style=flat-square)
+[!Supported Python Versions](https://img.shields.io/pypi/pyversions/tamr-unify-client.svg?style=flat-square)
```
pcattori, line 7: Is this intentional? I don't think the new version will result in a badge displayed on the README.

```suggestion
![Supported Python Versions](https://img.shields.io/pypi/pyversions/tamr-unify-client.svg?style=flat-square)
```

tamr_client/dataset/_dataset.py
```diff
@@ -262,5 +262,5 @@ def create(
     data = response.successful(r).json()
     dataset_path = data["relativeId"]
     dataset_url = URL(instance=instance, path=str(dataset_path))
-
+    # There may be a slight perf hit from using a GET to construct the dataset but this is OK
```
pcattori, line 265: I think this sort of comment belongs more in the `Contributor Docs` that inline given that we use this pattern (i.e. "Don't be fancy about avoiding extra requests until we identify a perf issue") throughout the codebase.

tamr_client/dataset/dataframe.py
```diff
@@ -60,3 +64,122 @@ def upsert(
         {primary_key_name: pk, **json.loads(row)} for pk, row in serialized_records
     )
     return record.upsert(session, dataset, records, primary_key_name=primary_key_name)
+
+
+def create_dataset(
```
pcattori, line 69: Given that `dataframe.upsert` already exists, should we be consistent and name this just `create`?

```suggestion
def create(
```

pcattori (changes requested):

README.md
skalish: Ah, this was a mistake on my part.  Correcting something that wasn't wrong in the first place.

tamr_client/dataset/dataframe.py
skalish: This was intended to avoid ambiguity over whether it is a dataset or a dataframe that is being created.  Given that there is no function to create a dataframe this is not really an issue but I'm not 100% sure here.

tamr_client/dataset/_dataset.py
skalish: Does this seem like an ADR-level note, or does it belong somewhere else in the Contributor Docs?

tamr_client/dataset/dataframe.py
skalish: I've convinced myself. The namespacing makes sense with `create`.  You can either create a dataset with `dataset.create` (read "dataset: create") or from a dataframe with `dataset.dataframe.create` (read: "dataset: dataframe-create").

tamr_client/dataset/_dataset.py
pcattori: Yea, actually ADR seems like the right place. Good point.

tamr_client/dataset/dataframe.py
```diff
@@ -60,3 +64,122 @@ def upsert(
...
+            _handle_creation_failure(session, ds, "Some records had validation errors")
+    except (TamrClientException, requests.HTTPError) as e:
+        _handle_creation_failure(session, ds, f"Record could not be created: {e}")
+
+    # Get Dataset from server
+    return dataset._dataset._by_url(session, ds.url)
+
+
+def _handle_creation_failure(session: Session, stub_dataset: Dataset, error: str):
+    """Attempt to make dataframe.create_dataset atomic by deleting the created dataset in the event
```
pcattori, line 147:
```suggestion
    """Attempt to make dataframe.create atomic by deleting the created dataset in the event
```

pcattori (changes requested):

pcattori (approved):

pcattori:
@skalish : can you rebase this off of `master` to bring it up-to-date?

PR #486, opened 2021-02-05
ci(release): do not use pip to install poetry

pcattori:
# ↪️ Pull Request

The poetry docs explicitly warn against using pip for installation. See
https://python-poetry.org/docs/#installing-with-pip. Instead, we'll use a Github Action tailor-made
for installing poetry.


skalish (approved):
LGTM. Only one way to find out if it works...

PR #490, opened 2021-05-19
Use session auth token cookie to minimize re-auth events

### LINKED ISSUE #305: Support use of session token after authenticating using username / password
# 🙋 feature request

Right now, there's no easy way to graduate from UsernamePasswordAuth to TokenAuth. The downside of just using UsernamePasswordAuth is twofold:
1. It creates a new session with every request, significantly increasing the load on the auth service
2. It re-sends credentials with every request, significantly increasing the exposure of the credentials.

The downside of TokenAuth is:
1. There is no way to get a session token, other than raw API calls.

Ideally, when creating a client with UsernamePasswordAuth, it would capture the returned session token and use the cached token to authenticate requests until the session times out.

## 🤔 Expected Behavior

```
tamr = api.Client(UsernamePasswordAuth(user, password), host=host)`
projects = client.projects()
datasets = client.datasets()
```
In the above example, projects and datasets should be fetched using the same session

## 😯 Current Behavior

In the above example, projects and datasets are fetched using different sessions.

## 💁 Possible Solution

Create an authenticator that takes the protocol, host, port, username, password of the auth service to fetch a session token, then behaves like token auth as long as the session token is valid.
Ideally, we'd then deprecate then remove `UsernamePasswordAuth` to encourage migration to the better pattern.

## 🔦 Context

Just try'n'a get stuff done, and I'm accidentally hammering on the auth service.

## 💻 Examples

See above.

skalish:
# ↪️ Pull Request

This PR adds the capability to fetch and store Tamr auth cookies to reuse sessions and limit the necessary number of calls to the Tamr auth service.

For TC, auth cookie fetching is done automatically whenever a request is made to the Tamr server and a valid cookie is not already stored.

For TUC, an auth cookie is fetched when a `Client` is created with the `store_auth_cookie` parameter set to `True`. It is `False` by default.  An existing `Client` without an auth cookie can be manually directed to fetch one with the method `.set_auth_cookie()`.

Closes #305 

## 💻 Examples

```python
from tamr_unify_client.client import Client

auth = UsernamePasswordAuth("my_username", "my_password")
client = Client(auth, store_auth_cookie=True)

# OR

client = Client(auth)
client.set_auth_cookie()
```

## ✔️ PR Todo

- [x] Testing for this change
- [x] Update relevant [docs](https://github.com/Datatamer/tamr-client/tree/main/docs) + docstrings

### DIFF
--- a/tamr_client/_types/session.py
+++ b/tamr_client/_types/session.py
@@ -1,3 +1,48 @@
+from typing import Optional
+
 import requests
 
-Session = requests.Session
+from tamr_client._types.auth import UsernamePasswordAuth
+
+
+class Session(requests.Session):
+    def __init__(self):
+        super(self.__class__, self).__init__()
+        self._stored_auth: Optional[UsernamePasswordAuth] = None
+
+    def request(self, *args, **kwargs):
+        # signature of `requests` requires not naming positional args
+        response = super(self.__class__, self).request(*args, **kwargs)
+        if response.status_code == 401 and "credentials" in response.text.lower():
+            first_response = response
+            self._set_auth_cookie(args[1])
+            response = super(self.__class__, self).request(*args, **kwargs)
+            if response.status_code == 401 and "credentials" in response.text.lower():
+                # Login credentials are bad, return original response
+                response = first_response
+        return response
+
+    def _set_auth_cookie(self, parent_url: str):
+        """Fetch and store an auth token for the given client configuration"""
+        # No-op if _stored_auth is None
+        if self._stored_auth is None:
+            return
+
+        # Fetch auth token and store as cookie
+        socket_address = parent_url.split("/api/")[0]
+        r = self.post(
+            socket_address + "/api/versioned/v1/instance:login",
+            json={
+                "username": self._stored_auth.username,
+                "password": self._stored_auth.password,
+            },
+        )
+        if r.status_code == 200:
+            auth_token = r.json()["token"]
+            self.cookies.set("authToken", auth_token)  # TODO: Set domain for security
+            # Clear session auth if cookie is retrieved
+            self.auth = None
+        else:
+            # Set session auth from client auth in case it has been cleared
+            # Allow the following call to pass credentials in header
+            self.auth = self._stored_auth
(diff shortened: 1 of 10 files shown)


pcattori:
What about something like:

```python
session: tc.Session = tc.session.make(instance=i, auth=a)
```

where `tc.Session` is a `dataclass`:

```python
@dataclass(frozen=true)
class Session:
  _session: requests.Session
  auth: requests.AuthThingy
  instance: tc.Instance
```

pcattori:
Or with monkey-patching:

```python
  def __init__(self):
    self._session.request = our request function?
```

tamr_unify_client/client.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         port: Optional[int] = 9100,
         base_path: str = "/api/versioned/v1/",
         session: Optional[requests.Session] = None,
+        store_auth_cookie: bool = False,  # TODO: Remove this - stopgap to pass existing tests
```
skalish, line 50: @pcattori What might be a better solution here?

pcattori: This sort of duct-tape solution is fine for TUC since its underlying implementation makes this sort of thing hard anyway.

tamr_client/_types/session.py
```diff
@@ -1,3 +1,42 @@
 import requests
 
-Session = requests.Session
+from tamr_client._types.auth import UsernamePasswordAuth
+
+
+class Session(requests.Session):
+    def __init__(self):
+        super(self.__class__, self).__init__()
+        self._stored_auth: UsernamePasswordAuth = UsernamePasswordAuth("", "")
```
pcattori, line 9: `UsernamePasswordAuth("", "")` is not an auth we ever expect to be valid. Could we change this type to be `Optional` and default to `None`? I think that more clearly captures that the session is not authorized by default.

skalish: This was originally to satisfy `mypy`, but I believe code changes since then should make a `None` default possible.  Testing locally now.

skalish: Done!

pcattori (approved):

PR #491, opened 2021-05-20
Allow hitting API endpoints with scheme-like relative urls

skalish:
# ↪️ Pull Request

Addresses an error that occurs when trying to reach the versioned API endpoints `instance:login` and `projects:import`.  Since these look like schemes, the function `urljoin` assumes they are the absolute URI.  An additional, and not ideal, change has been made to still accommodate making requests with full URIs, so long as they have the same protocol (`http`, `https`, etc.) as is configured for the client instance.

## 💻 Examples
```python
from tamr_unify_client import client
TAMR = client.Client(...)
TAMR.post("projects:import", json={...})
```
Before this PR, the underlying request would try to hit `projects:import` and fail with a `requests.exceptions.InvalidSchema` exception.  With this PR the request will succeed.

An alternative to this PR is to provide the guidance that the API endpoints of the format `abc:xyz` must be worked around by formatting with a `./`-prefix (`./abc:xyz`) or hit with and absolute URI.

## ✔️ PR Todo

- [x] Testing for this change

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -90,11 +90,12 @@ class Client:
             HTTP response from the Tamr server
         """
         # imperfect solution needed to allow passing absolute urls
-        if f"{self.host}:" in endpoint or endpoint[0] == "/":
-            url = urljoin(self.origin + self.base_path, endpoint)
-        else:
-            # prefix endpoint with "./" to handle endpoints with a colon and no other slashes
+        if endpoint.count(":") == 1 and "/" not in endpoint:
+            # prefix endpoint with "./" to handle endpoints with a colon and no slashes
             url = urljoin(self.origin + self.base_path, "./" + endpoint)
+        else:
+            # can handle with standard urljoin
+            url = urljoin(self.origin + self.base_path, endpoint)
         response = self.session.request(method, url, **kwargs)
 
         logger.info(
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -1,6 +1,6 @@
 import logging
 from typing import Optional
-from urllib.parse import urljoin
+from urllib.parse import urlparse
 
 import requests
 import requests.auth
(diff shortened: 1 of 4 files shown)


tamr_unify_client/client.py
```diff
@@ -89,7 +89,13 @@ def request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
...
             HTTP response from the Tamr server
         """
-        url = urljoin(self.origin + self.base_path, endpoint)
+        # imperfect solution needed to allow passing absolute urls
+        if endpoint.count(":") == 1 and "/" not in endpoint:
+            # prefix endpoint with "./" to handle endpoints with a colon and no slashes
+            url = urljoin(self.origin + self.base_path, "./" + endpoint)
+        else:
+            # can handle with standard urljoin
+            url = urljoin(self.origin + self.base_path, endpoint)
```
pcattori, line 98: The only benefit of `urljoin` over just string concatenation is for handling absolute URLs. Maybe we should just replace this whole block with:

```suggestion
        if endpoint.startswith('/'):
            url = self.origin + endpoint
        else:
            url = self.origin + self.base_path + endpoint
```

Note: ^I forget if we have to add a `/` in between `self.base_path` and `endpoint`

skalish: > The only benefit of `urljoin` over just string concatenation is for handling absolute URLs. Maybe we should just replace this whole block with:
> 
> Note: ^I forget if we have to add a `/` in between `self.base_path` and `endpoint`

It doesn't need the added `/`.  The only problem with this is that it no longer allows passing the full URI, e.g. 
`endpoint = http://localhost:9100/api/versioned/v1/instance:login` -> 
`url = http://localhost:9100/api/versioned/v1/http://localhost:9100/api/versioned/v1/instance:login`
We have tests around this that fail when I made the change.

pcattori: We could add a simple `if endpoint.startswith('http')` check as well

skalish: That sounds pretty good.  My main concern was whether there are any cases where `http` and `https` are not the first part of the url, but I'm unfamiliar with this.

pcattori: To be more robust, we could use `urllib.parse` to inspect if there is a scheme https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

skalish: This is where the API naming gets us into trouble, `urllib.parse.urlparse` will detect a scheme in the string "instance:login".

skalish: Maybe we need to check for both a schema and a netloc?

pcattori: How do we know that a customer didn't actually implement a protocol called `instance` or `projects`?

We either need a mechanism to unambiguously differentiate between any arbitrary protocol and URL path segment OR we need to make assumptions about what the scheme (and other parts like netloc) look like.

skalish: I made a change so that we look for any non-empty scheme and netloc, which I think does waht we need.  It uncovered another detail, which is that `urljoin` was also helping us sanitize urls occasionally. So if the endpoint was `projects/1/attributeConfigurations//1`, the `urljoin` call would remove the redundant second `/`.  One of our tests mistakenly caused this to happen, so I made a small change to that here too.
`

pcattori (approved):

PR #492, opened 2021-05-26
Project attributes

skalish:
# ↪️ Pull Request
Current functionality only supports accessing attributes of datasets (including as unified datasets), but projects have attributes as well.  The Tamr server requires attributes be created with project attribute endpoints rather than unified dataset attribute endpoints.

This PR updates type hinting to support the use of `Project` instances as the `parent` argument in attribute-fetching functions, and adds a check to prevent the creation of attributes in a unified dataset when creation must be done in the related project.

Also adds function `project.attributes(...)` to return all attributes of a project.

Open questions:
- The `ValueError` raised when trying to create an attribute for a unified dataset should be changed to a custom exception, but I'm not sure what to call it.
- This check is only made in the `attribute.create(...)` function.  Maybe it should be moved into the `_create` function and also implemented somehow in the `update`/`delete` functions.

## 💻 Examples

```python
import tamr_client as tc

s = tc.Session(...)
instance = tc.Instance(...)

project = tc.MasteringProject(...)
ud = tc.dataset.unified.from_project(s, project)  # has type UnifiedDataset

# Will fail
attr = tc.attribute.create(s, ud, name="my_attr", is_nullable=True)
>>> ValueError("Attributes for unified datasets must be created as attributes of the containing project")

# Will succeed
attr = tc.attribute.create(s, project, name="my_attr", is_nullable=True)

```

## ✔️ PR Todo

- [ ] Testing for this change

### DIFF
--- a/tamr_client/attribute/_attribute.py
+++ b/tamr_client/attribute/_attribute.py
@@ -3,10 +3,19 @@ See https://docs.tamr.com/reference/attribute-types
 """
 from copy import deepcopy
 from dataclasses import replace
-from typing import Optional
+from typing import Optional, Tuple, Union
 
 from tamr_client import response
-from tamr_client._types import Attribute, AttributeType, Dataset, JsonDict, Session, URL
+from tamr_client._types import (
+    Attribute,
+    AttributeType,
+    Dataset,
+    JsonDict,
+    Project,
+    Session,
+    UnifiedDataset,
+    URL,
+)
 from tamr_client.attribute import type as attribute_type
 from tamr_client.exception import TamrClientException
 
@@ -29,6 +38,8 @@ _RESERVED_NAMES = frozenset(
     ]
 )
 
+Parent = Union[Dataset, Project]
+
 
 class AlreadyExists(TamrClientException):
     """Raised when trying to create an attribute that already exists on the server"""
@@ -50,13 +61,21 @@ class ReservedName(TamrClientException):
     pass
 
 
-def by_resource_id(session: Session, dataset: Dataset, id: str) -> Attribute:
+class CannotCreateAttributesOnUnifiedDataset(TamrClientException):
+    """Raised when attempting to create an attribute on a unified dataset"""
+
+    pass
+
+
+def by_resource_id(
+    session: Session, parent: Union[Dataset, Project], id: str
+) -> Attribute:
     """Get attribute by resource ID
 
     Fetches attribute from Tamr server
 
     Args:
-        dataset: Dataset containing this attribute
+        parent: Dataset or project containing this attribute
         id: Attribute ID
 
     Raises:
@@ -64,7 +83,7 @@ def by_resource_id(session: Session, dataset: Dataset, id: str) -> Attribute:
             Corresponds to a 404 HTTP error.
         requests.HTTPError: If any other HTTP error is encountered.
     """
-    url = replace(dataset.url, path=dataset.url.path + f"/attributes/{id}")
+    url = replace(parent.url, path=parent.url.path + f"/attributes/{id}")
     return _by_url(session, url)
 
 
@@ -126,7 +145,7 @@ def to_json(attr: Attribute) -> JsonDict:
 
 def create(
     session: Session,
-    dataset: Dataset,
+    parent: Parent,
     *,
     name: str,
     is_nullable: bool,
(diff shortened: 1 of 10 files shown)


tamr_client/attribute/_attribute.py
```diff
@@ -50,21 +59,23 @@ class ReservedName(TamrClientException):
     pass
 
 
-def by_resource_id(session: Session, dataset: Dataset, id: str) -> Attribute:
+def by_resource_id(
+    session: Session, parent: Union[Dataset, Project], id: str
```
pcattori, line 63: You could define a type alias at the top of this file like `Parent = Union[Dataset, Project]` and then use `Parent` in the function signatures, just to consolidate the definition of the parent type.

tamr_client/attribute/_attribute.py
```diff
@@ -157,9 +168,15 @@ def create(
     if name in _RESERVED_NAMES:
         raise ReservedName(name)
 
+    if isinstance(parent, UnifiedDataset):
+        raise ValueError(
```
pcattori, line 172: For the exception name, I think explicitness is better than concision. So something like `raise CannotCreateAttributesOnUnifiedDataset` is fine by me.

tamr_client/project.py
```diff
@@ -198,3 +208,28 @@ def get_all(
...
+    r = session.get(str(attrs_url))
+    attrs_json = response.successful(r).json()
+
+    attrs = []
+    for attr_json in attrs_json:
+        id = attr_json["name"]
+        attr_url = replace(attrs_url, path=attrs_url.path + f"/{id}")
+        attr = _attribute_from_json(attr_url, attr_json)
+        attrs.append(attr)
+    return tuple(attrs)
```
pcattori, line 235: To avoid mutation logic and type conversion, you could define a helper function and use a tuple comprehension or `map` call:

```suggestion
    def attr_from_json(attr_json):
        id = attr_json["name"]
        attr_url = replace(attrs_url, path=attrs_url.path + f"/{id}")
        return _attribute_from_json(attr_url, attr_json)
    return tuple(map(attr_from_json, attrs_json))
```

skalish: The `project.attributes` and `dataset.attributes` functions only differ in their typehints and docstrings.  They are logically identical.  Is there a better practice than having this repeated code, or is the clarity of having them defined separately more valuable?

pcattori: I think its a question of whether the repeated logic has an intuitive tag/name. If yes, make a shared utility function with that name. If no and the shared logic is small, its not too bad to keep some minor duplication.

skalish: This case does make sense to have a shared utility function `attribute._get_all_from_parent`, so I've done that.

tamr_client/attribute/_attribute.py
skalish: @pcattori What do you think, is this a reasonable level of protection to build in?

Currently, `<project_id>` is always an integer, but in the broadest possible generality, the suggested solution could break on "/"-containing project IDs (if that's even possible).

pcattori: Is there a way to refer to attributes via a canonical URL that is not relative to a project nor a UD?

If no: we can do some rudimentary checking of the URL as you suggest. Probably worth doing.
If yes: we'd have do some gymnastics to figure out if its a UA or a project attribute. Probably not worth doing, too complex.

skalish: There are internal API endpoints that allow specifying a unified attribute by an ID alone, e.g. `343`, but there is no way of identifying an attribute through a versioned URL except for
1. `/v1/datasets/{datasetId}/attributes/{attributeId}`
2. `/v1/projects/{project}/unifiedDataset/attributes/{attributeId}` (an alias of the above)
3. `/v1/projects/{project}/attributes/{attributeId}`
where 2 and 3 are only possible for project/unified dataset attributes.

Using tamr_client as currently designed, it is possible to get a unified attribute `Attribute`, with a standard dataset URL that we can't detect with the simple check, if the user gets the unified dataset by resource ID or name as a regular `Dataset` and then gets the attribute from there.  The simple check only works if the user gets the unified dataset using the `tamr_client.dataset.unified.from_project` function, vs the `tamr_client.dataset.by_resource_id` or `.by_name` functions.

pcattori: Let's skip on doing guardrails here. Let's do the simple thing, get it out there, and then act on any bug reports with more context.

pcattori (approved):

PR #494, opened 2021-06-30
fix(tuc): change set_auth_cookie to defer to legacy functionality on request errors

skalish:
# ↪️ Pull Request

Fixes an oversight in #490 that could give confusing error messages when e.g. using incorrect credentials

## ✔️ PR Todo

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -144,8 +144,10 @@ class Client:
             "./instance:login",
             json={"username": self.auth.username, "password": self.auth.password},
         )
-        if r.status_code != 200:
-            raise Exception  # TODO: Replace with sensible handling
+        # If login request fails for any reason do not set cookie so following responses use
+        # header credentials for authentication
+        if not r.ok:
+            return
         auth_token = r.json()["token"]
         self.session.cookies.set("authToken", auth_token)  # TODO: Set domain
         # Clear session auth.  It is still accessible as self.auth if needed


tamr_unify_client/client.py
```diff
@@ -144,12 +144,12 @@ def set_auth_cookie(self):
...
             json={"username": self.auth.username, "password": self.auth.password},
         )
-        if r.status_code != 200:
-            raise Exception  # TODO: Replace with sensible handling
-        auth_token = r.json()["token"]
-        self.session.cookies.set("authToken", auth_token)  # TODO: Set domain
-        # Clear session auth.  It is still accessible as self.auth if needed
-        self.session.auth = None
+        # Only continue if request is successful
+        if r.ok:
```
pcattori, line 148: alternative: I slightly prefer a guard clause here. up to you 😄 

```suggestion
        if !r.ok: return
```

less indentation and keeps it clear what the "happy path" is

skalish: I'm good with that.  For some reason my brain wasn't happy with adding a return to a method, but that's obviously fine.

tamr_unify_client/client.py
```diff
@@ -144,8 +144,9 @@ def set_auth_cookie(self):
             "./instance:login",
             json={"username": self.auth.username, "password": self.auth.password},
         )
-        if r.status_code != 200:
-            raise Exception  # TODO: Replace with sensible handling
+        # Only continue if request is successful
```
pcattori, line 147: nit: I think this comment doesn't say anything not said by the code. I'd recommend removing it or replacing with something that explains _why_ we are doing this conceptually if you think its worth explaining.

pcattori (approved):

pcattori (approved):

PR #496, opened 2021-12-01
Add Python 3.9 to test matrix

skalish:
Also bump patch versions in `.python-version`

# ↪️ Pull Request

## ✔️ PR Todo

- [ ] Testing for this change


skalish:
Seems like the issue of hanging install is fixed by running `poetry update` locally to bump all dependencies in `poetry.lock` to newer versions.  Any thoughts on whether this is a good or bad idea @pcattori?

skalish:
> > Seems like the issue of hanging install is fixed by running `poetry update` locally to bump all dependencies in `poetry.lock` to newer versions. Any thoughts on whether this is a good or bad idea @pcattori?
> 
> Seems fine to me as long as our CI test suite passes with the updated versions. To be clear: are you proposing we run `poetry.update` locally and commit the changes to `poetry.lock` so that CI picks that up? Or are you suggesting we run `poetry.update` within the CI script? (I think you mean the former, in which case 👍 )

Yep, running `poetry.update` and commiting the new `poetry.lock`, which I think constitutes a "chore".  Please correct if not.

PR #498, opened 2022-10-26
SUP-6198: Add support for JWT authentication

### LINKED ISSUE #497: Support JWT Authentication
# 🙋 feature request

Tamr Core now supports [Authentication with JSON Web Tokens (JWT)](https://docs.tamr.com/new/reference/authentication#authentication-with-json-web-tokens-jwt) and so we should too.

skalish:
# ↪️ Pull Request

This PR adds JWT token authentication for both `tamr_client` and `tamr_unify_client`

Closes #497.

## ✔️ PR Todo

- [ ] Testing for this change

### DIFF
--- a/tamr_unify_client/client.py
+++ b/tamr_unify_client/client.py
@@ -6,7 +6,6 @@ import requests
 import requests.auth
 import requests.exceptions
 
-from tamr_unify_client.auth.jwt_token import JwtTokenAuth
 from tamr_unify_client.auth.username_password import UsernamePasswordAuth
 from tamr_unify_client.dataset.collection import DatasetCollection
 from tamr_unify_client.project.collection import ProjectCollection
@@ -106,7 +105,7 @@ class Client:
         if (
             response.status_code == 401
             and "credentials" in response.text.lower()
-            and not isinstance(self.session.auth, JwtTokenAuth)
+            and isinstance(self.auth, UsernamePasswordAuth)
         ):
             first_response = response
             self.set_auth_cookie()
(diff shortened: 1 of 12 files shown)


sammaximtamr (approved):
This looks good to me! Nice comments and tests in this. We should wait for @Andy-L 's review though.

tamr_unify_client/client.py
```diff
@@ -102,7 +103,11 @@ def request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
 
         # Attempt request with auth cookie
         response = self.session.request(method, url, **kwargs)
-        if response.status_code == 401 and "credentials" in response.text.lower():
+        if (
+            response.status_code == 401
+            and "credentials" in response.text.lower()
+            and not isinstance(self.session.auth, JwtTokenAuth)
```
Andy-L, line 109: if `set_auth_cookie()` only works for `UsernamePasswordAuth`, then it would make more sense to check here that you are using `UsernamePasswordAuth`, instead of the negative check for `JwtTokenAuth`

Andy-L (approved):
Looks ok to me, though I haven't written any python in a while.
It looks like there are 2 separate client implementations, is that right?
Just wondering why we need 2 implementations of the JWT auth.

skalish:
> Looks ok to me, though I haven't written any python in a while. It looks like there are 2 separate client implementations, is that right? Just wondering why we need 2 implementations of the JWT auth.

There are two client implementations because this library contains two packages:
- the object-oriented `tamr_unify_client` implementation
- the functional `tamr_client`, formerly considered s BETA but now released

The auth and session implementation is pretty similar in both of them, so it seemed reasonable to update both to support JWT here.

PR #500, opened 2023-05-16
SUP-6396: Pin version of urllib3

skalish:
# ↪️ Pull Request
This PR pins the version of the transitive dependency `urllib3` (required by `requests`) to use versions before 2.0.0.

It also updates the tested versions of Python and upgrades some dev dependencies.

## ✔️ PR Todo
- [x] Update relevant [docs](https://github.com/Datatamer/tamr-client/tree/main/docs) + docstrings


.python-version
```diff
@@ -1,4 +1,4 @@
-3.6.13
+
 3.7.10
 3.8.10
 3.9.5
```
clint-richardson: do you not need to add a 3.10 here? I'm reviewing each commit so apologies if this is addressed in a later one

skalish: This should be addressed in a later commit. These are not strictly necessary at all but do help point to specific patch versions for local development.

stubs/responses.pyi
```diff
@@ -15,9 +15,6 @@ def add(
     status: Optional[int] = None,
     json: Optional[JsonDict] = None,
 ): ...
-
-T = TypeVar("T")
-
-def activate(T) -> T: ...
+def activate(Callable) -> Callable: ...
```
clint-richardson, line 18: what is this change doing? Is it related to updating the versions of python and urllib?

README.md
```diff
@@ -38,5 +38,4 @@ For more see the [official docs](https://tamr-client.readthedocs.io/en/stable/).
 
 ## Maintainers
```
clint-richardson, line 40: @nbateshaus would you be comfortable with us adding you to this list?

stubs/responses.pyi
skalish: When updating the version of the `mypy` typechecker, it started complaining about the previous type-hint.  We use this only as a decorator of test functions for handling response mocking. This satisfies it and also seems reasonable.

clint-richardson (approved):

